PR #6, opened 2016-08-19
Remove intermediary buffer in decode_mode

dyule:
I wanted to get rid of the `buffer` variable in `decode_mode`, since it was an unnecessary memory allocation, but I got a little overexcited and reworked the method to use entirely iterators instead.

If you have any stylistic or naming concerns, I am happy to address them.


alicemaz:
sorry to have neglected this, I was squeamish about the iterators for fussy style reasons and let it slip my mind. believe this is handled by #9 now

PR #8, opened 2016-12-08
Update lib.rs

tredoe:
Related to [issue 7](https://github.com/alicemaz/rust-base64/issues/7)


alicemaz:
lgtm. need to fix the import but I will just do that after merging. thanks!

PR #9, opened 2016-12-30
Allow encoding/decoding into user-supplied buffers

marshallpierce:
While evaluating my base64 options, I liked the look of this library, but allocation makes me sad. So, given your note in Goals about wanting to support an allocation-free mode, I thought I'd take a stab at it.

You'll find the following:

- `encode_mode_buf` and `decode_mode_buf` variants that take a user-supplied mutable `String` or `Vec<u8>` as appropriate
- Benchmarks for encoding and decoding in various sizes, with and without reusing buffers
- More tests with exhaustive "every possible byte" tests for very short inputs and randomized longer inputs
- New decode logic that's about 6x faster. Not [AVX2 fast](https://github.com/aklomp/base64) but pretty fast...

More on that last one: once I had benchmarks up, it was pretty clear that decoding was quite slow: around 100MiB/s on my (fast) system. I changed the decode logic to use a lookup table and some manual unrolling (similar to how encoding is done). This more than doubled the instructions per clock to around 3.3, almost entirely eliminated branch predictor misses, and got around a 5-6x speedup. However, the new decode loop **does not support padding in the middle of the message** and will return `InvalidByte` if it finds such padding. That's not backwards compatible with the previous implementation. Hopefully the speedup is worth it... (I've never actually seen such base64 in practice, FWIW.) It is possible to shoehorn support for moving out of the fast loop until internal padding is consumed and then going back into fast mode later, but it seemed like a lot of complexity for something that people don't seem to do.

Below are some benchmark results I got on my i7-6850K. I didn't do a super careful job of isolating the system (turning off all the power-saving automatic downclocking, etc) so there is some run-to-run variance but the overall trends should be clear. Note that `cargo bench` does a bad job of guessing how many iterations to run for the first benchmark, so that one is especially dodgy.

Benchmarks on original encoding/decoding algorithms, but with optional encode buffer reuse (from 99603c167a1acacdb6df510fde5167e94c81add9):

```
test decode_100b            ... bench:         379 ns/iter (+/- 21)
test decode_10mib           ... bench: 104,987,833 ns/iter (+/- 800,636)
test decode_30mib           ... bench: 314,499,514 ns/iter (+/- 1,053,282)
test decode_3b              ... bench:          39 ns/iter (+/- 2)
test decode_3kib            ... bench:      23,618 ns/iter (+/- 2,075)
test decode_3mib            ... bench:  31,418,189 ns/iter (+/- 460,024)
test decode_500b            ... bench:       1,793 ns/iter (+/- 76)
test decode_50b             ... bench:         207 ns/iter (+/- 12)
test encode_100b            ... bench:         256 ns/iter (+/- 0)
test encode_100b_reuse_buf  ... bench:         214 ns/iter (+/- 0)
test encode_10mib           ... bench:  25,927,177 ns/iter (+/- 191,033)
test encode_10mib_reuse_buf ... bench:  22,972,252 ns/iter (+/- 66,665)
test encode_30mib           ... bench:  77,459,890 ns/iter (+/- 350,972)
test encode_30mib_reuse_buf ... bench:  68,810,891 ns/iter (+/- 86,976)
test encode_3b              ... bench:          25 ns/iter (+/- 0)
test encode_3b_reuse_buf    ... bench:          10 ns/iter (+/- 0)
test encode_3kib            ... bench:       6,534 ns/iter (+/- 5)
test encode_3kib_reuse_buf  ... bench:       6,508 ns/iter (+/- 18)
test encode_3mib            ... bench:   7,478,577 ns/iter (+/- 146,395)
test encode_3mib_reuse_buf  ... bench:   6,761,072 ns/iter (+/- 25,675)
test encode_500b            ... bench:       1,111 ns/iter (+/- 2)
test encode_500b_reuse_buf  ... bench:       1,056 ns/iter (+/- 0)
test encode_50b             ... bench:         150 ns/iter (+/- 0)
test encode_50b_reuse_buf   ... bench:         105 ns/iter (+/- 0)
```

Decoding is 4x slower than encoding. 

With the new decoding logic and optional decode buffer reuse (5409ef88300f494b8b466a2244d06a5cf856d4e9) *-- updated to fix decode benchmark buffer size calculation*:

```
test decode_100b            ... bench:         143 ns/iter (+/- 9) = 699 MB/s
test decode_100b_reuse_buf  ... bench:         122 ns/iter (+/- 3) = 819 MB/s
test decode_10mib           ... bench:  13,929,090 ns/iter (+/- 1,875,204) = 752 MB/s
test decode_10mib_reuse_buf ... bench:  12,931,825 ns/iter (+/- 565,734) = 810 MB/s
test decode_30mib           ... bench:  43,703,792 ns/iter (+/- 468,047) = 719 MB/s
test decode_30mib_reuse_buf ... bench:  38,776,028 ns/iter (+/- 785,088) = 811 MB/s
test decode_3b              ... bench:          37 ns/iter (+/- 0) = 108 MB/s
test decode_3b_reuse_buf    ... bench:          14 ns/iter (+/- 14) = 285 MB/s
test decode_3kib            ... bench:       3,683 ns/iter (+/- 8) = 834 MB/s
test decode_3kib_reuse_buf  ... bench:       3,688 ns/iter (+/- 8) = 832 MB/s
test decode_3mib            ... bench:   4,205,466 ns/iter (+/- 51,438) = 748 MB/s
test decode_3mib_reuse_buf  ... bench:   3,809,695 ns/iter (+/- 11,031) = 825 MB/s
test decode_500b            ... bench:         627 ns/iter (+/- 1) = 797 MB/s
test decode_500b_reuse_buf  ... bench:         607 ns/iter (+/- 2) = 823 MB/s
test decode_50b             ... bench:          92 ns/iter (+/- 0) = 565 MB/s
test decode_50b_reuse_buf   ... bench:          66 ns/iter (+/- 1) = 787 MB/s
test encode_100b            ... bench:         256 ns/iter (+/- 0) = 390 MB/s
test encode_100b_reuse_buf  ... bench:         212 ns/iter (+/- 0) = 471 MB/s
test encode_10mib           ... bench:  25,947,829 ns/iter (+/- 137,031) = 404 MB/s
test encode_10mib_reuse_buf ... bench:  22,959,165 ns/iter (+/- 43,985) = 456 MB/s
test encode_30mib           ... bench:  78,035,153 ns/iter (+/- 757,898) = 403 MB/s
test encode_30mib_reuse_buf ... bench:  68,851,707 ns/iter (+/- 97,494) = 456 MB/s
test encode_3b              ... bench:          25 ns/iter (+/- 0) = 120 MB/s
test encode_3b_reuse_buf    ... bench:           9 ns/iter (+/- 0) = 333 MB/s
test encode_3kib            ... bench:       6,533 ns/iter (+/- 17) = 470 MB/s
test encode_3kib_reuse_buf  ... bench:       6,505 ns/iter (+/- 13) = 472 MB/s
test encode_3mib            ... bench:   7,492,017 ns/iter (+/- 130,224) = 419 MB/s
test encode_3mib_reuse_buf  ... bench:   6,757,431 ns/iter (+/- 27,139) = 465 MB/s
test encode_500b            ... bench:       1,112 ns/iter (+/- 3) = 449 MB/s
test encode_500b_reuse_buf  ... bench:       1,056 ns/iter (+/- 1) = 473 MB/s
test encode_50b             ... bench:         150 ns/iter (+/- 0) = 333 MB/s
test encode_50b_reuse_buf   ... bench:         105 ns/iter (+/- 0) = 476 MB/s
```

Now decoding is significantly faster than encoding.


seanmonstar:
Drive-by Bonus: if you set `b.bytes = size as u64;`, the benchmark results will output the bytes per second for you.

marshallpierce:
Nice, thanks @seanmonstar !

Keats:
@marshallpierce would you mind me piggy-backing on your PR to fix #10 ?

marshallpierce:
@Keats I don't mind piggy-backing, but I don't think [RFC 4648](https://tools.ietf.org/html/rfc4648#section-5) agrees that the URL safe alphabet also implies the lack of padding:

>    The pad character "=" is typically percent-encoded when used in an
   URI [9], but if the data length is known implicitly, this can be
   avoided by skipping the padding; see section 3.2.

I read that as saying "the default would be to have padding, but you can skip it if it makes sense in your circumstance", so I would say that a general-purpose library like this should probably emit the `=`. Now, a separate question is: should we add additional surface to the API to let the user choose whether to emit `=`? Or should we maybe just say "use `.trim_right_matches('=')` if you don't want padding".

BTW, decoding (for this library at least) already works with or without padding, and by glancing at rustc_serialize's decoding logic it looks like they should support that too.

Keats:
> BTW, decoding (for this library at least) already works with or without padding, and by glancing at rustc_serialize's decoding logic it looks like they should support that too.
Yeah it does, I was testing the wrong thing without realising :(

Regarding configs:
Golang does something a bit like that https://golang.org/pkg/encoding/base64/#pkg-variables and `rustc-serialize` allows defining custom configs too https://github.com/rust-lang-nursery/rustc-serialize/blob/master/src/base64.rs#L57-L58
I'd go for the same approach as `rustc-serialize` myself as the crate is overall quite nice to use (including impl `from_base64` and `to_base64` traits) so it might be worth copying a bit

marshallpierce:
FWIW I think the config approach is reasonable. I've used rustc_serialize's base64 in other projects and found the config system pretty sane.

Keats:
I'd wait a bit for @alicemaz input as well as the current PR + config combined would be pretty much a full rewrite of the crate

alicemaz:
terribly sorry, just got off a long vacation. will take a closer look soon, but thank you for all this!

I'm fine with erroring out on interspered '=' characters, I allowed it since it was a "may" in rfc4648 and figured why not, but given the speedup that is enough of a why not for me

thanks again!

marshallpierce:
Great, glad the increased error strictness isn't a problem. More speedups to come. You can see progress at https://gist.github.com/marshallpierce/20351d44b29e238fecb821dc87251e5f.

marshallpierce:
OK, I'm calling decoding good. Decode performance is now 80% as fast as the aklomp's C compiled with gcc and 90% as fast as C compiled with clang (a more fair comparison since rustc uses LLVM too). I've tried various tweaks from here with no consistently measurable benefit. If this approach looks good, I'll port the same ideas to encoding.

```
test decode_100b            ... bench:          96 ns/iter (+/- 5) = 1041 MB/s
test decode_100b_reuse_buf  ... bench:          75 ns/iter (+/- 1) = 1333 MB/s
test decode_10mib           ... bench:   8,754,671 ns/iter (+/- 274,348) = 1197 MB/s
test decode_10mib_reuse_buf ... bench:   7,334,820 ns/iter (+/- 129,159) = 1429 MB/s
test decode_30mib           ... bench:  26,617,411 ns/iter (+/- 622,087) = 1181 MB/s
test decode_30mib_reuse_buf ... bench:  22,387,310 ns/iter (+/- 182,809) = 1405 MB/s
test decode_3b              ... bench:          42 ns/iter (+/- 2) = 95 MB/s
test decode_3b_reuse_buf    ... bench:          19 ns/iter (+/- 1) = 210 MB/s
test decode_3kib            ... bench:       2,004 ns/iter (+/- 37) = 1532 MB/s
test decode_3kib_reuse_buf  ... bench:       1,976 ns/iter (+/- 40) = 1554 MB/s
test decode_3mib            ... bench:   2,536,762 ns/iter (+/- 217,774) = 1240 MB/s
test decode_3mib_reuse_buf  ... bench:   2,114,590 ns/iter (+/- 23,871) = 1487 MB/s
test decode_500b            ... bench:         352 ns/iter (+/- 10) = 1420 MB/s
test decode_500b_reuse_buf  ... bench:         341 ns/iter (+/- 11) = 1466 MB/s
test decode_50b             ... bench:          67 ns/iter (+/- 0) = 776 MB/s
test decode_50b_reuse_buf   ... bench:          46 ns/iter (+/- 0) = 1130 MB/s
```

tests/tests.rs
```diff
@@ -44,25 +158,259 @@ fn decode_rfc4648_4() {
...
 
+
 //same
 #[test]
-fn decode_allow_absurd_pad() {
-    compare_decode("alice", "==Y=Wx===pY=2U=====");
+fn decode_absurd_pad_error() {
+    match decode("==Y=Wx===pY=2U=====").unwrap_err() {
+        Base64Error::InvalidByte(size, byte) => {
+            assert_eq!(0, size);
```
marshallpierce: Now forbidden, was allowed

tests/tests.rs
```diff
@@ -44,25 +158,259 @@ fn decode_rfc4648_4() {
...
-    compare_decode("alice", "YWxpY2U=====");
+fn decode_pad_inside_fast_loop_chunk_error() {
+    // can't PartialEq Base64Error, so we do this the hard way
+    match decode("YWxpY2U=====").unwrap_err() {
+        Base64Error::InvalidByte(offset, byte) => {
+            // since the first 8 bytes are handled in the fast loop, the
+            // padding is an error. Could argue that the *next* padding
+            // byte is technically the first erroneous one, but reporting
+            // that accurately is more complex and probably nobody cares
+            assert_eq!(7, offset);
```
marshallpierce: Slightly janky error detection when there's so much extra padding that it spills into the fast loop

tests/tests.rs
```diff
@@ -44,25 +158,259 @@ fn decode_rfc4648_4() {
...
+        _ => assert!(false)
+    }
+}
+
+#[test]
+fn decode_extra_pad_after_fast_loop_chunk_error() {
+    match decode("YWxpY2UABB===").unwrap_err() {
+        Base64Error::InvalidByte(offset, byte) => {
+            // extraneous third padding byte
+            assert_eq!(12, offset);
```
marshallpierce: Rejects extra padding at the end

src/lib.rs
```diff
@@ -204,59 +250,224 @@ pub fn encode_mode(bytes: &[u8], mode: Base64Mode) -> String {
...
+            // 2 - Padding after zero or one non-padding characters before it
+            //     in the current quad.
+            // 3 - More than two characters of padding. If 3 or 4 padding chars
+            //     are in the same quad, that implies it will be caught by #2.
+            //     If it spreads from one quad to another, it will be caught by
+            //     #2 in the second quad.
+
+            if i % 4 < 2 {
+                // Check for case #2.
+                // TODO InvalidPadding error
```
marshallpierce: Something to ponder: This isn't really an "invalid byte" in the since of illegal input to decode, it's that padding is wrong. Should we have a separate `InvalidPadding`? Pro: we *can* detect this error. Con: Probably nobody will ever care. 

I'm leaning towards "don't bother" at the moment but maybe this will solve somebody's longstanding problem!

src/lib.rs
```diff
@@ -204,59 +250,224 @@ pub fn encode_mode(bytes: &[u8], mode: Base64Mode) -> String {
...
+        // '=' padding
+        if *b == 0x3D {
+            // There can be bad padding in a few ways:
+            // 1 - Padding with non-padding characters after it
+            // 2 - Padding after zero or one non-padding characters before it
+            //     in the current quad.
+            // 3 - More than two characters of padding. If 3 or 4 padding chars
+            //     are in the same quad, that implies it will be caught by #2.
+            //     If it spreads from one quad to another, it will be caught by
+            //     #2 in the second quad.
```
marshallpierce: This is much stricter padding checking than what was enforced before. Now you can't really do too much, but I view that as a positive. I think you can probably still have just 1 padding char when you might have used 2, but that's about it.

marshallpierce: I added a test and that is still allowed (having `=` when you could use `==` to complete the 4-byte quad).

Keats:
@alicemaz are you ok with going for a config approach like rustc-serialize does?

alicemaz:
@marshallpierce looks good to me! will merge this now, couple other things to bring in before doing a release. thank you again!

@Keats I like rustc-serialize's config/traits (and had been intending to copy their approach when adding MIME, since that would require more options there didn't seem a reason to lock them to just three specific combinations). allowing encoding to not be emitted is fine, though I disagree with rustc-serialize that it should be the default for url-safe and read the rfc similarly to @marshallpierce on that point. we should be decoding the same regardless of the presence or absence of padding

marshallpierce:
thanks @alicemaz! I'll follow up with improvments to encoding along the same lines, hopefully without such a meandering path through optimization-space. :)

PR #11, opened 2017-01-17
Add config struct instead of mode

Keats:
Based on rustc-serialize, minus the MIME
specific fields


Keats:
Unrelated but I'd also like to do a PR for `ToBase64` and `FromBase64` traits that `rustc-serialize` have, so you can do `my_string.to_base64(base64::STANDARD)`, would you be open to that PR?

Keats:
@alicemaz ping when you have time

PR #14, opened 2017-02-13
Add {From,To}Base64 traits for convenience

Keats:
Code mostly taken from rustc_serialize

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,6 @@
 extern crate byteorder;
 
-use std::{fmt, error, str};
+use std::{fmt, error, str, string};
 
 use byteorder::{BigEndian, ByteOrder};
 
@@ -28,10 +28,10 @@ pub static STANDARD: Config = Config {char_set: CharacterSet::Standard, pad: tru
 pub static URL_SAFE: Config = Config {char_set: CharacterSet::UrlSafe, pad: true};
 pub static URL_SAFE_NO_PAD: Config = Config {char_set: CharacterSet::UrlSafe, pad: false};
 
-
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug)]
 pub enum Base64Error {
     Utf8(str::Utf8Error),
+    FromUtf8(string::FromUtf8Error),
     InvalidByte(usize, u8),
     InvalidLength,
 }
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -28,10 +28,10 @@ pub static STANDARD: Config = Config {char_set: CharacterSet::Standard, pad: tru
 pub static URL_SAFE: Config = Config {char_set: CharacterSet::UrlSafe, pad: true};
 pub static URL_SAFE_NO_PAD: Config = Config {char_set: CharacterSet::UrlSafe, pad: false};
 
-
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug)]
 pub enum Base64Error {
     Utf8(str::Utf8Error),
+    FromUtf8(string::FromUtf8Error),
```
marshallpierce: I'm not sure this is a good idea... Now, someone who's not using the traits still has to handle this error case even though it's impossible in their usage, which isn't ideal. TBH I don't find these sorts of convenience traits all that convenient as they tend to create project-specific API of things like String based on what traits are mixed on, and that inhibits new developers coming in to a project and things like that, but I don't mind them as long as they're opt-in. (Also, they encourage the allocation-heavy flavor of base64 usage, which makes me wipe away a tiny tear.) However, the presence of this error variant means that now everyone has to care/know about the extra traits even when they don't actually use them.

In fact, I'm pretty sure we can ditch the existing `Utf8` error variant as well since we create (correct) UTF-8 via the powers of unsafe. ;)

It looks like this is only even needed because of `FromBase64` for `[u8]`. Rather than having the intervening string and UTF-8, we should just expose a way to decode bytes. (It already is just handling bytes under the hood, anyway.) Hmm... perhaps `decode_config_buf` should have its input be bytes? Or have another similar function, though we already have so many `decode_*`. WDYT?

Keats: I actually agree that it isn't necessarily needed (and I don't use it myself) but I thought it was quite nice to have in terms of UX.
I'm not sure another `decode_` method is worth it, we can probably close down that PR

alicemaz:
is the consensus that we no longer want this because of the error situation (and now that ut8 error is excised)?

incidentally #18 made the decodes polymorphic, I don't see any reason not to do the same with encodes? I think this would accomplish the objective of more flexibility without complicating the api

marshallpierce:
I lean towards not having this. It steers people towards less efficient code (not reusing buffers), and it's at best arguable if it makes any sense to syntactically treat things like base64 encoding/decoding as an intrinsic property of bytes, strings, etc. If a project finds that for some reason they prefer being able to call .to_base_64() or something like that, they can always have their own trait in a couple of lines of code.

The only downside to decoding into a slice rather than a vec is that it raises the possibility of the slice not being long enough, so we'd need a new error variant to express that, but we could also have slice-style and Vec-style decode methods with separate error enums so that each style could have an error enum that precisely expressed the possible error cases.

Keats:
We can close it yes.

PR #18, opened 2017-03-25
Allow decode functions to decode a &[u8] (fixes #16)

### LINKED ISSUE #16: Should `decode_*` methods support decoding from a `&[u8]?`
See https://github.com/alicemaz/rust-base64/pull/14/files#diff-b4aea3e418ccdb71239b96952d9cddb6R130

It seems unfortunate to force a user who has some (ascii) bytes that they wish to decode to `String`-ify them first (at the cost of verifying that the bytes are UTF8, or unsafely creating the string). In the decode logic we simply access the bytes of the string anyway, so it would be little work to expose decoding from bytes.

We already check for every possible invalid byte because the decode tables have 256 entries (with suitable invalid markers), so it wouldn't incur any additional checking beyond what we already have. Thoughts?

nox:
This is not a breaking change.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -103,7 +103,7 @@ pub fn encode(input: &[u8]) -> String {
 ///    println!("{:?}", bytes);
 ///}
 ///```
-pub fn decode(input: &str) -> Result<Vec<u8>, Base64Error> {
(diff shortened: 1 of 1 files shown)


alicemaz:
ty!

nox:
Thanks for the minor release!

Your crate is now in Servo's tree. :)

https://github.com/servo/servo/pull/16137

PR #19, opened 2017-03-26
Add MIME support

alicemaz:
MIME config option, LineEnding enum, add necessary code to encode and decode
also get rid of decode_ws

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -21,51 +21,70 @@ pub enum LineEnding {
     CRLF,
 }
 
+#[derive(Clone, Copy, Debug)]
+pub enum LineWrap {
+    NoWrap,
+    Wrap(usize, LineEnding)
+}
+
 /// Contains configuration parameters for base64 encoding
 #[derive(Clone, Copy, Debug)]
 pub struct Config {
     /// Character set to use
-    pub char_set: CharacterSet,
+    char_set: CharacterSet,
     /// True to pad output with `=` characters
-    pub pad: bool,
+    pad: bool,
     /// Remove whitespace before decoding, at the cost of an allocation
-    pub strip_whitespace: bool,
-    /// Characters per line, None (or Some(0), but less "proper") for no linebreaks
-    pub line_size: Option<usize>,
-    /// Unix or Windows line endings, ignored if above None/Some(0)
-    pub line_ending: LineEnding,
+    strip_whitespace: bool,
+    /// ADT signifying whether to linewrap output, and if so by how many characters and with what ending
+    line_wrap: LineWrap,
+}
+
+impl Config {
+    pub fn new(char_set: CharacterSet,
+               pad: bool,
+               strip_whitespace: bool,
+               input_line_wrap: LineWrap) -> Config {
+        let line_wrap = match input_line_wrap  {
+            LineWrap::Wrap(0, _) => LineWrap::NoWrap,
+            _ => input_line_wrap,
+        };
+
+        Config {
+            char_set: char_set,
+            pad: pad,
+            strip_whitespace: strip_whitespace,
+            line_wrap: line_wrap,
+        }
+    }
 }
 
 pub static STANDARD: Config = Config {
     char_set: CharacterSet::Standard,
     pad: true,
     strip_whitespace: false,
-    line_size: None,
-    line_ending: LineEnding::LF,
+    line_wrap: LineWrap::NoWrap,
 };
 
 pub static MIME: Config = Config {
     char_set: CharacterSet::Standard,
     pad: true,
     strip_whitespace: true,
-    line_size: Some(76),
-    line_ending: LineEnding::CRLF,
+    line_wrap: LineWrap::Wrap(76, LineEnding::CRLF),
 };
 
 pub static URL_SAFE: Config = Config {
     char_set: CharacterSet::UrlSafe,
     pad: true,
     strip_whitespace: false,
-    line_size: None,
-    line_ending: LineEnding::LF,
+    line_wrap: LineWrap::NoWrap,
 };
 
 pub static URL_SAFE_NO_PAD: Config = Config {
     char_set: CharacterSet::UrlSafe,
     pad: false,
     strip_whitespace: false,
-    line_size: None,
-    line_ending: LineEnding::LF,
+    line_wrap: LineWrap::NoWrap,
 };
 
 
@@ -170,13 +189,10 @@ fn encoded_size(bytes_len: usize, config: Config) -> usize {
     } else {
         complete_output_chars + 4
     };
-    let line_ending_length = match config.line_ending {
-        LineEnding::CRLF => 2,
-        LineEnding::LF => 1,
-    };
-    let line_ending_output_chars = match config.line_size {
-        Some(0) | None => 0,
-        Some(n) => printing_output_chars / n * line_ending_length,
+    let line_ending_output_chars = match config.line_wrap {
+        LineWrap::NoWrap => 0,
+        LineWrap::Wrap(n, LineEnding::CRLF) => printing_output_chars / n * 2,
+        LineWrap::Wrap(n, LineEnding::LF) => printing_output_chars / n,
     };
 
     return printing_output_chars + line_ending_output_chars;
@@ -280,15 +296,14 @@ pub fn encode_config_buf(input: &[u8], config: Config, buf: &mut String) {
         }
     }
 
-    if config.line_size.is_some() && config.line_size.unwrap() > 0 {
-        let line_size = config.line_size.unwrap();
+    if let LineWrap::Wrap(line_size, line_end) = config.line_wrap {
         let len = raw.len();
         let mut i = 0;
         let mut j = 0;
 
         while i < len {
             if i > 0 && i % line_size == 0 {
-                match config.line_ending {
+                match line_end {
                     LineEnding::LF => { raw.insert(j, b'\n'); j += 1; }
                     LineEnding::CRLF => { raw.insert(j, b'\r'); raw.insert(j + 1, b'\n'); j += 2; }
                 }
@@ -556,7 +571,7 @@ pub fn decode_config_buf<T: ?Sized + AsRef<[u8]>>(input: &T,
 
 #[cfg(test)]
 mod tests {
-    use super::{encoded_size, STANDARD, MIME, Config, CharacterSet, LineEnding};
+    use super::*;
 
     #[test]
     fn encoded_size_correct() {
@@ -610,13 +625,12 @@ mod tests {
 
     #[test]
     fn encoded_size_correct_lf() {
-        let config = Config {
-            char_set: CharacterSet::Standard,
-            pad: true,
-            strip_whitespace: false,
-            line_size: Some(76),
-            line_ending: LineEnding::LF,
-        };
+        let config = Config::new(
+            CharacterSet::Standard,
+            true,
+            false,
+            LineWrap::Wrap(76, LineEnding::LF)
+        );
 
         assert_eq!(0, encoded_size(0, config));
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,6 +15,12 @@ pub enum CharacterSet {
     UrlSafe
 }
 
+#[derive(Clone, Copy, Debug)]
+pub enum LineEnding {
+    LF,
+    CRLF,
+}
+
 /// Contains configuration parameters for base64 encoding
 #[derive(Clone, Copy, Debug)]
 pub struct Config {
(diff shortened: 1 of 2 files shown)


alicemaz:
will have a chance to update readme in a few days, gonna leave this open/unmerged till then. adds support for encoding with \n or \r\n line endings every n characters, and decoding without regard to whitespace through the normal config struct, a superset of the functionality required for RFC 2045/MIME. so as not to un-fast the fast loops, stripping whitespace for encoding requires an allocation, and adding linebreaks for decoding requires another trip through the buffer and expensive inserts. however doing it this way means existing uses do not incur a performance penalty

master:
```running 32 tests
test decode_100b            ... bench:         148 ns/iter (+/- 3) = 675 MB/s
test decode_100b_reuse_buf  ... bench:         120 ns/iter (+/- 1) = 833 MB/s
test decode_10mib           ... bench:  13,768,882 ns/iter (+/- 157,466) = 761 MB/s
test decode_10mib_reuse_buf ... bench:  11,485,179 ns/iter (+/- 115,779) = 912 MB/s
test decode_30mib           ... bench:  41,568,400 ns/iter (+/- 299,614) = 756 MB/s
test decode_30mib_reuse_buf ... bench:  34,851,858 ns/iter (+/- 276,693) = 902 MB/s
test decode_3b              ... bench:          52 ns/iter (+/- 0) = 76 MB/s 
test decode_3b_reuse_buf    ... bench:          27 ns/iter (+/- 0) = 148 MB/s
test decode_3kib            ... bench:       3,011 ns/iter (+/- 28) = 1020 MB/s
test decode_3kib_reuse_buf  ... bench:       2,999 ns/iter (+/- 31) = 1024 MB/s
test decode_3mib            ... bench:   3,893,603 ns/iter (+/- 42,231) = 807 MB/s
test decode_3mib_reuse_buf  ... bench:   3,258,866 ns/iter (+/- 18,864) = 965 MB/s
test decode_500b            ... bench:         535 ns/iter (+/- 19) = 934 MB/s
test decode_500b_reuse_buf  ... bench:         511 ns/iter (+/- 4) = 978 MB/s
test decode_50b             ... bench:          94 ns/iter (+/- 1) = 553 MB/s
test decode_50b_reuse_buf   ... bench:          70 ns/iter (+/- 0) = 742 MB/s
test encode_100b            ... bench:         125 ns/iter (+/- 2) = 800 MB/s
test encode_100b_reuse_buf  ... bench:          97 ns/iter (+/- 1) = 1030 MB/s
test encode_10mib           ... bench:  15,152,112 ns/iter (+/- 267,088) = 692 MB/s
test encode_10mib_reuse_buf ... bench:  11,794,474 ns/iter (+/- 94,737) = 889 MB/s
test encode_30mib           ... bench:  45,870,731 ns/iter (+/- 373,142) = 685 MB/s
test encode_30mib_reuse_buf ... bench:  35,447,231 ns/iter (+/- 167,884) = 887 MB/s
test encode_3b              ... bench:          40 ns/iter (+/- 1) = 75 MB/s 
test encode_3b_reuse_buf    ... bench:          16 ns/iter (+/- 0) = 187 MB/s
test encode_3kib            ... bench:       2,516 ns/iter (+/- 25) = 1220 MB/s
test encode_3kib_reuse_buf  ... bench:       2,500 ns/iter (+/- 37) = 1228 MB/s
test encode_3mib            ... bench:   4,220,447 ns/iter (+/- 73,831) = 745 MB/s
test encode_3mib_reuse_buf  ... bench:   3,439,425 ns/iter (+/- 12,400) = 914 MB/s
test encode_500b            ... bench:         442 ns/iter (+/- 5) = 1131 MB/s
test encode_500b_reuse_buf  ... bench:         419 ns/iter (+/- 4) = 1193 MB/s
test encode_50b             ... bench:          78 ns/iter (+/- 1) = 641 MB/s
test encode_50b_reuse_buf   ... bench:          54 ns/iter (+/- 0) = 925 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 32 measured```

branch:
```running 32 tests
test decode_100b            ... bench:         150 ns/iter (+/- 2) = 666 MB/s
test decode_100b_reuse_buf  ... bench:         123 ns/iter (+/- 2) = 813 MB/s
test decode_10mib           ... bench:  13,530,090 ns/iter (+/- 204,946) = 774 MB/s
test decode_10mib_reuse_buf ... bench:  11,314,633 ns/iter (+/- 148,715) = 926 MB/s
test decode_30mib           ... bench:  40,771,624 ns/iter (+/- 1,022,577) = 771 MB/s
test decode_30mib_reuse_buf ... bench:  34,408,071 ns/iter (+/- 317,427) = 914 MB/s
test decode_3b              ... bench:          53 ns/iter (+/- 0) = 75 MB/s 
test decode_3b_reuse_buf    ... bench:          29 ns/iter (+/- 0) = 137 MB/s
test decode_3kib            ... bench:       2,983 ns/iter (+/- 66) = 1029 MB/s
test decode_3kib_reuse_buf  ... bench:       2,963 ns/iter (+/- 75) = 1036 MB/s
test decode_3mib            ... bench:   3,818,325 ns/iter (+/- 66,805) = 823 MB/s
test decode_3mib_reuse_buf  ... bench:   3,212,207 ns/iter (+/- 25,436) = 979 MB/s
test decode_500b            ... bench:         534 ns/iter (+/- 10) = 936 MB/s
test decode_500b_reuse_buf  ... bench:         505 ns/iter (+/- 4) = 990 MB/s
test decode_50b             ... bench:          97 ns/iter (+/- 2) = 536 MB/s
test decode_50b_reuse_buf   ... bench:          72 ns/iter (+/- 4) = 722 MB/s
test encode_100b            ... bench:         136 ns/iter (+/- 3) = 735 MB/s
test encode_100b_reuse_buf  ... bench:         103 ns/iter (+/- 3) = 970 MB/s
test encode_10mib           ... bench:  15,223,341 ns/iter (+/- 173,537) = 688 MB/s
test encode_10mib_reuse_buf ... bench:  12,051,608 ns/iter (+/- 111,463) = 870 MB/s
test encode_30mib           ... bench:  45,332,565 ns/iter (+/- 325,761) = 693 MB/s
test encode_30mib_reuse_buf ... bench:  36,157,098 ns/iter (+/- 233,365) = 870 MB/s
test encode_3b              ... bench:          49 ns/iter (+/- 1) = 61 MB/s 
test encode_3b_reuse_buf    ... bench:          19 ns/iter (+/- 0) = 157 MB/s
test encode_3kib            ... bench:       2,564 ns/iter (+/- 29) = 1198 MB/s
test encode_3kib_reuse_buf  ... bench:       2,528 ns/iter (+/- 21) = 1215 MB/s
test encode_3mib            ... bench:   4,240,298 ns/iter (+/- 111,968) = 741 MB/s
test encode_3mib_reuse_buf  ... bench:   3,497,667 ns/iter (+/- 33,489) = 899 MB/s
test encode_500b            ... bench:         460 ns/iter (+/- 6) = 1086 MB/s
test encode_500b_reuse_buf  ... bench:         428 ns/iter (+/- 6) = 1168 MB/s
test encode_50b             ... bench:          86 ns/iter (+/- 2) = 581 MB/s
test encode_50b_reuse_buf   ... bench:          57 ns/iter (+/- 1) = 877 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 32 measured```

File: src/lib.rs [hunk 1]
```diff
@@ -15,18 +15,58 @@ pub enum CharacterSet {
...
     /// Character set to use
     pub char_set: CharacterSet,
     /// True to pad output with `=` characters
     pub pad: bool,
+    /// Remove whitespace before decoding, at the cost of an allocation
+    pub strip_whitespace: bool,
+    /// Characters per line, None (or Some(0), but less "proper") for no linebreaks
+    pub line_size: Option<usize>,
+    /// Unix or Windows line endings, ignored if above None/Some(0)
+    pub line_ending: LineEnding,
```
marshallpierce: I think it might be nice to have the type system provide more guidance about the line wrap stuff, perhaps by having a dedicated `LineWrap` struct. There's basically an implicit `NoWrap` line wrap style that is applied by having line_size be 0 or `None`. Leaking that detail to the outside world means we might end up in a compatibility pickle as we make more changes to Config. If we hid that detail from the world, we could have more freedom to change that later, and also avoid user confusion about the sometimes-ignored `line_ending` member. So, something like:

```
struct LineWrap {
    line_size: Option<usize>,
    line_ending: LineEnding
}

impl struct LineWrap {
    pub fn no_wrap() -> LineWrap {
        LineWrap {line_size: None, line_ending: LF}
    }

    pub fn wrap(line_size: usize, line_ending: LineEnding) -> LineWrap {
        // check line_size > 0
        ...
    }
}
```

Now we can hide the details of how `line_ending` is applied, and the only two valid use cases are explicitly labelled with those methods.

A second thought... This `Config` struct is getting a little beefy. `char_set` is meaningful for both encoding and decoding, but the others are only used for either encoding or decoding, not both.

WDYT about a `EncodeConfig` with `char_set`, `pad`, and `line_wrap`, and a `DecodeConfig` with `char-set` and `strip_whitespace`?

src/lib.rs
```diff
@@ -255,6 +279,25 @@ pub fn encode_config_buf(input: &[u8], config: Config, buf: &mut String) {
             raw.push(0x3d);
         }
     }
+
+    if config.line_size.is_some() && config.line_size.unwrap() > 0 {
```
marshallpierce: This could use `config.line_size.unwrap_or(0) > 0`, or even store that in a local to avoid the unwrap below

src/lib.rs
```diff
@@ -255,6 +279,25 @@ pub fn encode_config_buf(input: &[u8], config: Config, buf: &mut String) {
...
+        let line_size = config.line_size.unwrap();
+        let len = raw.len();
+        let mut i = 0;
+        let mut j = 0;
+
+        while i < len {
+            if i > 0 && i % line_size == 0 {
+                match config.line_ending {
+                    LineEnding::LF => { raw.insert(j, b'\n'); j += 1; }
+                    LineEnding::CRLF => { raw.insert(j, b'\r'); raw.insert(j + 1, b'\n'); j += 2; }
```
marshallpierce: This would be quadratic in the output length, right? It might not be worth the effort since I can't think of a case where line-wrapped base64 is used for large volumes of data, but a linear-time way of doing this would be to encode smaller chunks of input to get to at least the desired line length worth of encoded text, insert line endings as needed, save the leftovers for the next line, repeat. Every once in a while the leftovers themselves will probably get to be a line of their own, depending on the specific line length. However, I'm not convinced that for the likely use case (decoding a PEM, I imagine) it's really worth it. Also, people who need speed could always hand-roll that around the lower-level logic.

src/lib.rs
```diff
@@ -303,7 +346,16 @@ pub fn decode_config_buf<T: ?Sized + AsRef<[u8]>>(input: &T,
                                                   config: Config,
                                                   buffer: &mut Vec<u8>)
                                                   -> Result<(), DecodeError> {
-    let input_bytes = input.as_ref();
+    let mut input_copy;
+    let input_bytes = if config.strip_whitespace {
+        input_copy = Vec::<u8>::with_capacity(input.as_ref().len());
+        input_copy.extend(input.as_ref().iter().filter(|b| !b" \n\t\r\x0c".contains(b)));
```
marshallpierce: We could allocate less here if we only used an intermediate buffer of, say, 4KiB and repeatedly filled it with non-whitespace chars and decoded that into the output buffer. Again, not sure it's really worth the effort since most whitespacey base64 is probably PEM files that are (1) small and (2) read once on startup, and we're not closing the door on people implementing their own faster logic if they need it.

src/lib.rs
alicemaz: was actually going to change to
```
pub enum LineWrap {
    NoWrap, 
    Wrap(usize, LineEnding)
}
```
to be more succinct, but a private struct could work too (also an ADT doesn't solve the potential confusion with 0, tfw no dependent types)

tbh I think just having the one config struct is fine for now. I'd honestly be surprised if many people even used it, I'd expect 90% of usecases to be encode/decode with implicit STANDARD and most of the rest using one of the other predefined ones

src/lib.rs
alicemaz: ya it would be. alternative simple solution could just allocate another vec and copy in linear

src/lib.rs
alicemaz: could also make Wrap imply whitespace-ignoring decode and remove the bool from the config struct but this is probably not intuitive

alicemaz:
@marshallpierce making the config fields private and giving the whole thing a single constructor seemed more reasonable than filling fields one with a special constructor for a subobject. I just have it implicitly convert 0-size wrap to nowrap, though we could just as easily return a result and make this an error

xmo-odoo:
@marshallpierce does this still work or has it been dropped? `Config` doesn't seem to have any sort of wrapping configuration, so RFCs 1421, 2045, and 4880 outputs don't seem supported (anymore?)

marshallpierce:
See https://crates.io/crates/line-wrap for all your line wrapping needs.

xmo-odoo:
> See [crates.io/crates/line-wrap](https://crates.io/crates/line-wrap) for all your line wrapping needs.

nice :+1:

PR #26, opened 2017-04-30
Add a few fuzzers. Fixes #21

### LINKED ISSUE #21: Add support for cargo-fuzz
https://github.com/rust-fuzz/cargo-fuzz

Seems like a nice way to stress test encoding and decoding random data. We have some unit tests that do random data, but it'd be nice to just leave this running for, say, an hour...

marshallpierce:
I had done this as part of working on line wrapping and a Display wrapper, but that PR was getting enormous so I'm splitting up into parts. This is the first, to be followed by line wrapping, then Display.

### DIFF
--- a/fuzz/fuzzers/utils.rs
+++ b/fuzz/fuzzers/utils.rs
@@ -1,0 +1,48 @@
+extern crate base64;
+extern crate rand;
+extern crate ring;
+
+use self::base64::*;
+use self::rand::{Rng, SeedableRng, XorShiftRng};
+use self::rand::distributions::{IndependentSample, Range};
+use self::ring::digest;
+
+pub fn random_config(data: &[u8]) -> Config {
+    // use sha256 of data as rng seed so it's repeatable
+    let sha = digest::digest(&digest::SHA256, data);
+    let sha_bytes = sha.as_ref();
+
+    let mut seed = [0; 4];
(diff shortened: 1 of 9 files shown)


alicemaz:
I have not had a chance to try this one out, and I'm leaving for lambdaconf in a day sans computer. I assume though it is just test-related stuff, feel free to merge if you please

fuzz/fuzzers/utils.rs
```diff
@@ -0,0 +1,48 @@
...
+extern crate ring;
+
+use self::base64::*;
+use self::rand::{Rng, SeedableRng, XorShiftRng};
+use self::rand::distributions::{IndependentSample, Range};
+use self::ring::digest;
+
+pub fn random_config(data: &[u8]) -> Config {
+    // use sha256 of data as rng seed so it's repeatable
+    let sha = digest::digest(&digest::SHA256, data);
```
elsid, line 12: Could a more lightweight dependency like [sha2](https://github.com/RustCrypto/hashes/blob/457061f19b3538651aa7ef208ce6b04bfba65e61/sha2/Cargo.toml) be used here instead of `ring`?

marshallpierce: I don't see why not.

PR #27, opened 2017-04-30
Line wrapping optimization. Fixes #25

### LINKED ISSUE #25: line seperation behaves badly for non-empty buffers
(for now I wrapped it in `if orig_buf_len == 0` and published it as a patch so it doesn't mess up anyone's day)

so this is a thing that hadn't occurred to me at the time. since it just runs through the whole buffer adding line endings every n chars, data that's already there gets mutated, most often in an extremely undesirable manner. for instance mime-encoding a string that produces an output > 76, then encoding emptystring into the same buffer adds extra line seps

not sure the best way to handle this. the naive solution is to start counting from the starting point of the input, which _could_ make sense if the user were using a large string for assorted data (though the fact that we unconditionally write to the end would make this cumbersome, and anyway there are so much more barriers in the way of this compared to buffer sharing to dodge mallocs in C I don't see why one would do it in rust)

more likely I assume the user would be incrementally building a single item, in which case we would want to count from the _previous_ line sep, so as to always produce n-character lines with the final line unterminated (eg user produces 80 chars, we break after 76 for a length of 82, they add 80 more and we break again at 76*2+2 for 164). but making these kinds of assumptions about what already is in the buffer rather than encoding them into types makes the haskeller in me queasy and start wanting to implement yet another wrapper over strings

(incidental to this, the reserve calculation is wrong in the case where a non-empty buffer would be large enough to add even more line-endings, though only wrong in a "oops it doubles in capacity" and doesn't affect the fast loop)

marshallpierce:
Reimplements line wrapping to use a linear algorithm instead of a quadratic one. By 3 KiB, it was 17x faster (see benchmarks below). I didn't check in a 1MiB benchmark because it took *forever* to run on the old code, but it's 690e3 ns after optimization and 1e9 before, about 1400x faster. 

Initial commit (only added benchmarks, namely the `_mime` ones):

```
test encode_100b                ... bench:          89 ns/iter (+/- 0) = 1123 MB/s
test encode_100b_reuse_buf      ... bench:          71 ns/iter (+/- 0) = 1408 MB/s
test encode_10mib               ... bench:   9,008,847 ns/iter (+/- 308,063) = 1163 MB/s
test encode_10mib_reuse_buf     ... bench:   5,681,881 ns/iter (+/- 9,253) = 1845 MB/s
test encode_30mib               ... bench:  27,901,401 ns/iter (+/- 165,404) = 1127 MB/s
test encode_30mib_reuse_buf     ... bench:  17,047,398 ns/iter (+/- 1,735,299) = 1845 MB/s
test encode_3b                  ... bench:          32 ns/iter (+/- 0) = 93 MB/s
test encode_3b_reuse_buf        ... bench:          13 ns/iter (+/- 0) = 230 MB/s
test encode_3kib                ... bench:       1,707 ns/iter (+/- 3) = 1799 MB/s
test encode_3kib_reuse_buf      ... bench:       1,689 ns/iter (+/- 4) = 1818 MB/s
test encode_3kib_reuse_buf_mime ... bench:      33,256 ns/iter (+/- 66) = 92 MB/s
test encode_3mib                ... bench:   2,454,545 ns/iter (+/- 26,395) = 1281 MB/s
test encode_3mib_reuse_buf      ... bench:   1,693,286 ns/iter (+/- 4,663) = 1857 MB/s
test encode_500b                ... bench:         309 ns/iter (+/- 2) = 1618 MB/s
test encode_500b_reuse_buf      ... bench:         291 ns/iter (+/- 1) = 1718 MB/s
test encode_500b_reuse_buf_mime ... bench:       5,115 ns/iter (+/- 29) = 97 MB/s
test encode_50b                 ... bench:          59 ns/iter (+/- 0) = 847 MB/s
test encode_50b_reuse_buf       ... bench:          41 ns/iter (+/- 0) = 1219 MB/s
```

After optimization:

```
test encode_100b                ... bench:          84 ns/iter (+/- 0) = 1190 MB/s
test encode_100b_reuse_buf      ... bench:          70 ns/iter (+/- 0) = 1428 MB/s
test encode_10mib               ... bench:   9,088,328 ns/iter (+/- 132,041) = 1153 MB/s
test encode_10mib_reuse_buf     ... bench:   5,686,001 ns/iter (+/- 41,927) = 1844 MB/s
test encode_30mib               ... bench:  28,056,341 ns/iter (+/- 158,054) = 1121 MB/s
test encode_30mib_reuse_buf     ... bench:  17,019,695 ns/iter (+/- 20,809) = 1848 MB/s
test encode_3b                  ... bench:          27 ns/iter (+/- 0) = 111 MB/s
test encode_3b_reuse_buf        ... bench:          12 ns/iter (+/- 0) = 250 MB/s
test encode_3kib                ... bench:       1,699 ns/iter (+/- 6) = 1808 MB/s
test encode_3kib_reuse_buf      ... bench:       1,683 ns/iter (+/- 5) = 1825 MB/s
test encode_3kib_reuse_buf_mime ... bench:       1,951 ns/iter (+/- 16) = 1574 MB/s
test encode_3mib                ... bench:   2,469,405 ns/iter (+/- 34,354) = 1273 MB/s
test encode_3mib_reuse_buf      ... bench:   1,690,765 ns/iter (+/- 4,259) = 1860 MB/s
test encode_500b                ... bench:         305 ns/iter (+/- 1) = 1639 MB/s
test encode_500b_reuse_buf      ... bench:         290 ns/iter (+/- 0) = 1724 MB/s
test encode_500b_reuse_buf_mime ... bench:         355 ns/iter (+/- 1) = 1408 MB/s
test encode_50b                 ... bench:          55 ns/iter (+/- 0) = 909 MB/s
test encode_50b_reuse_buf       ... bench:          40 ns/iter (+/- 0) = 1250 MB/s
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,10 +1,13 @@
 extern crate byteorder;
 
-use std::{fmt, error, str};
+use std::{fmt, error, ptr, str};
 
 use byteorder::{BigEndian, ByteOrder};
 
 mod tables;
+
+mod line_wrap;
+use line_wrap::{line_wrap_parameters, line_wrap};
 
 /// Available encoding character sets
 #[derive(Clone, Copy, Debug)]
@@ -15,15 +18,41 @@ pub enum CharacterSet {
     UrlSafe
 }
 
+impl CharacterSet {
+    fn encode_table(&self) -> &'static [u8; 64] {
+        match self {
+            &CharacterSet::Standard => tables::STANDARD_ENCODE,
+            &CharacterSet::UrlSafe => tables::URL_SAFE_ENCODE
+        }
+    }
+
+    fn decode_table(&self) -> &'static [u8; 256] {
+        match self {
+            &CharacterSet::Standard => tables::STANDARD_DECODE,
+            &CharacterSet::UrlSafe => tables::URL_SAFE_DECODE
+        }
+    }
+}
+
 #[derive(Clone, Copy, Debug)]
 pub enum LineEnding {
     LF,
     CRLF,
 }
 
+impl LineEnding {
+    fn len(&self) -> usize {
+        match self {
+            &LineEnding::LF => 1,
+            &LineEnding::CRLF => 2
+        }
+    }
+}
+
 #[derive(Clone, Copy, Debug)]
 pub enum LineWrap {
     NoWrap,
+    // wrap length is always > 0
     Wrap(usize, LineEnding)
 }
 
@@ -171,7 +200,7 @@ pub fn decode<T: ?Sized + AsRef<[u8]>>(input: &T) -> Result<Vec<u8>, DecodeError
 ///}
 ///```
 pub fn encode_config<T: ?Sized + AsRef<[u8]>>(input: &T, config: Config) -> String {
-    let mut buf = String::with_capacity(encoded_size(input.as_ref().len(), config));
+    let mut buf = String::with_capacity(encoded_size(input.as_ref().len(), &config));
 
     encode_config_buf(input, config, &mut buf);
 
@@ -179,23 +208,34 @@ pub fn encode_config<T: ?Sized + AsRef<[u8]>>(input: &T, config: Config) -> Stri
 }
 
 /// calculate the base64 encoded string size, including padding
-fn encoded_size(bytes_len: usize, config: Config) -> usize {
+fn encoded_size(bytes_len: usize, config: &Config) -> usize {
+    let msg = "Encoded size exceeds usize";
     let rem = bytes_len % 3;
 
     let complete_input_chunks = bytes_len / 3;
-    let complete_output_chars = complete_input_chunks * 4;
-    let printing_output_chars = if rem == 0 {
-        complete_output_chars
+    let complete_chunk_output = complete_input_chunks.checked_mul(4).expect(msg);
+
+    let encoded_len_no_wrap = if rem > 0 {
+        if config.pad {
+            complete_chunk_output.checked_add(4).expect(msg)
+        } else {
+            let encoded_rem = match rem {
+                1 => 2,
+                2 => 3,
+                _ => panic!("Impossible remainder")
+            };
+            complete_chunk_output.checked_add(encoded_rem).expect(msg)
+        }
     } else {
-        complete_output_chars + 4
-    };
-    let line_ending_output_chars = match config.line_wrap {
-        LineWrap::NoWrap => 0,
-        LineWrap::Wrap(n, LineEnding::CRLF) => printing_output_chars / n * 2,
-        LineWrap::Wrap(n, LineEnding::LF) => printing_output_chars / n,
+        complete_chunk_output
     };
 
-    return printing_output_chars + line_ending_output_chars;
+    match config.line_wrap {
+        LineWrap::NoWrap => encoded_len_no_wrap,
+        LineWrap::Wrap(line_len, line_ending) => {
+            line_wrap_parameters(encoded_len_no_wrap, line_len, line_ending).total_len
+        }
+    }
 }
 
 ///Encode arbitrary octets as base64.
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -15,15 +18,41 @@ pub enum CharacterSet {
     UrlSafe
 }
 
+impl CharacterSet {
+    fn encode_table(&self) -> &'static [u8; 64] {
```
marshallpierce: I did this because I ended up duplicating those `match`es in the alternate encoding logic for the `Display` wrapper, but I think I actually like it even before that logic makes it an (aesthetic) necessity.

src/lib.rs
```diff
@@ -15,15 +18,41 @@ pub enum CharacterSet {
...
+        match self {
+            &CharacterSet::Standard => tables::STANDARD_ENCODE,
+            &CharacterSet::UrlSafe => tables::URL_SAFE_ENCODE
+        }
+    }
+
+    fn decode_table(&self) -> &'static [u8; 256] {
```
marshallpierce: I moved this one in just to have symmetry with `encode_table`

src/lib.rs
```diff
@@ -171,31 +200,42 @@ pub fn decode<T: ?Sized + AsRef<[u8]>>(input: &T) -> Result<Vec<u8>, DecodeError
...
+    let mut buf = String::with_capacity(encoded_size(input.as_ref().len(), &config));
 
     encode_config_buf(input, config, &mut buf);
 
     buf
 }
 
 /// calculate the base64 encoded string size, including padding
-fn encoded_size(bytes_len: usize, config: Config) -> usize {
+fn encoded_size(bytes_len: usize, config: &Config) -> usize {
```
marshallpierce: No reason to be copying `Config` around internally, so I made it take a `&`

src/lib.rs
```diff
@@ -171,31 +200,42 @@ pub fn decode<T: ?Sized + AsRef<[u8]>>(input: &T) -> Result<Vec<u8>, DecodeError
...
+    let msg = "Encoded size exceeds usize";
     let rem = bytes_len % 3;
 
     let complete_input_chunks = bytes_len / 3;
-    let complete_output_chars = complete_input_chunks * 4;
-    let printing_output_chars = if rem == 0 {
-        complete_output_chars
+    let complete_chunk_output = complete_input_chunks.checked_mul(4).expect(msg);
```
marshallpierce: As Rust's support for 16 bit platforms broadens I'm trying to at least provide useful error messages rather than silently wrong results for tiny address space sizes.

src/lib.rs
```diff
@@ -218,105 +258,140 @@ fn encoded_size(bytes_len: usize, config: Config) -> usize {
...
+            add_padding(input_bytes.len(), &mut b64_output[b64_bytes_written..])
+        } else {
+            0
+        };
+
+        let wrappable_bytes = b64_bytes_written + padding_bytes;
+
+        let line_ending_bytes = match config.line_wrap {
+            LineWrap::Wrap(line_len, line_end) =>
+                line_wrap(b64_output, wrappable_bytes, line_len, line_end),
```
marshallpierce: because we're wrapping only the contents of `b64_output` (which excludes pre-existing contents), #25 is avoided.

src/lib.rs
```diff
@@ -218,105 +258,140 @@ fn encoded_size(bytes_len: usize, config: Config) -> usize {
...
-    while leftover_index < start_of_rem {
-        raw.push(charset[(input_bytes[leftover_index] >> 2) as usize]);
-        raw.push(charset[((input_bytes[leftover_index] << 4 | input_bytes[leftover_index + 1] >> 4) & 0x3f) as usize]);
-        raw.push(charset[((input_bytes[leftover_index + 1] << 2 | input_bytes[leftover_index + 2] >> 6) & 0x3f) as usize]);
-        raw.push(charset[(input_bytes[leftover_index + 2] & 0x3f) as usize]);
 
-        leftover_index += 3;
+    while input_index < start_of_rem {
+        unsafe {
+            ptr::write(output_ptr,
```
marshallpierce: This is probably undetectably faster than `push()` but I switched it to use `ptr::write` like the fast loop because I felt it was easier to read and understand when all the writes use the same function, indexing, etc.

Also, now that we write into a slice, `push()` isn't available!

src/lib.rs
```diff
@@ -218,105 +258,140 @@ fn encoded_size(bytes_len: usize, config: Config) -> usize {
...
-    // we're only going to insert valid utf8
-    let mut raw = unsafe { buf.as_mut_vec() };
-    // start at the first free part of the output buf
-    let mut output_ptr = unsafe { raw.as_mut_ptr().offset(orig_buf_len as isize) };
+        // write into the newly reserved space
+        let b64_bytes_written = encode_to_slice(input_bytes, b64_output,
+                                                    config.char_set.encode_table());
+
+        let padding_bytes = if config.pad {
+            add_padding(input_bytes.len(), &mut b64_output[b64_bytes_written..])
```
marshallpierce: We need to be able to encode separately from padding for the Display wrapper, but I think this reads well even without that requirement.

src/lib.rs
```diff
@@ -574,88 +646,4 @@ pub fn decode_config_buf<T: ?Sized + AsRef<[u8]>>(input: &T,
 }
 
 #[cfg(test)]
-mod tests {
-    use super::*;
```
marshallpierce: moved to `src/tests.rs`, with a few more asserts and also other test cases.

src/line_wrap.rs
```diff
@@ -0,0 +1,339 @@
...
+    let (num_lines_with_endings, last_line_length) = if input_len % line_len > 0 {
+        // Every full line has an ending since there is a partial line at the end
+        (input_len / line_len, input_len % line_len)
+    } else {
+        // Every line is a full line, but no trailing ending.
+        // Subtraction will not underflow since we know input_len > line_len.
+        (input_len / line_len - 1, line_len)
+    };
+
+    // TODO should we expose exceeding usize via Result to be kind to 16-bit users? Or is that
```
marshallpierce: Not sure how prevalent large base64'ing will be on a 16 bit platform. I just went through this with https://github.com/jonhoo/hdrsample and we now have some `Result`s in more places. Sometimes inconvenient, unless you're on AVR and want to know when you can safely base64 something...

src/line_wrap.rs
```diff
@@ -0,0 +1,339 @@
...
+
+
+/// Insert line endings into the encoded base64 after each complete line (except the last line, even
+/// if it is complete).
+/// The provided buffer must be large enough to handle the increased size after endings are
+/// inserted.
+/// `input_len` is the length of the encoded data in `encoded_buf`.
+/// `line_len` is the width without line ending characters.
+/// Returns the number of line ending bytes added.
+pub fn line_wrap(encoded_buf: &mut [u8], input_len: usize, line_len: usize, line_ending: LineEnding) -> usize {
```
marshallpierce: So much unsafe. :( Please review with suspicion!

src/tests.rs
```diff
@@ -0,0 +1,337 @@
...
+
+        encode_config_buf(&input_buf, config, &mut encoded_buf);
+
+        assert_encode_sanity(&encoded_buf, &config, input_len);
+    }
+}
+
+
+#[test]
+fn encode_into_nonempty_buffer_doesnt_clobber_existing_contents() {
```
marshallpierce: This test failed with the old line wrap logic, as we would expect given #25

tests/tests.rs
```diff
@@ -18,98 +18,47 @@ fn compare_decode_mime(expected: &str, target: &str) {
...
-fn push_rand(buf: &mut Vec<u8>, len: usize) {
-    let mut r = rand::weak_rng();
-
-    for _ in 0..len {
-        buf.push(r.gen::<u8>());
-    }
-}
-
 // generate every possible byte string recursively and test encode/decode roundtrip
-fn roundtrip_append_recurse(byte_buf: &mut Vec<u8>, str_buf: &mut String, remaining_bytes: usize) {
```
marshallpierce: This got simplified along with its `_padding` variant into one function that takes a `Config`.

tests/tests.rs
```diff
@@ -18,98 +18,47 @@ fn compare_decode_mime(expected: &str, target: &str) {
...
+            assert_eq!(input_buf, decode_buf);
         }
 
-        byte_buf.truncate(orig_length);
+        input_buf.truncate(orig_length);
     }
 }
 
 // generate random contents of the specified length and test encode/decode roundtrip
-fn roundtrip_random(byte_buf: &mut Vec<u8>, str_buf: &mut String, byte_len: usize,
```
marshallpierce: Same for this one.

tests/tests.rs
```diff
@@ -431,100 +334,100 @@ fn roundtrip_random_with_fast_loop_no_padding() {
...
 #[test]
 fn roundtrip_random_4_byte() {
     let mut byte_buf: Vec<u8> = Vec::new();
     let mut str_buf = String::new();
 
-    roundtrip_random(&mut byte_buf, &mut str_buf, 4, 48, 10000);
+    roundtrip_random(&mut byte_buf, &mut str_buf, STANDARD, 4, 48, 10000);
 }
 
-//TODO like, write a thing to test every ascii val lol
```
marshallpierce: I think we've addressed this TODO by now. :)

alicemaz:
sorry I broke the merge, had to push a security patch. I will fixup and start reviewing this tomorrow

marshallpierce:
No problem, I merged in your changes and added a few other `checked_add()`'s. Most of the line wrapping logic was already pretty laden with `checked_*` fortunately.

marshallpierce:
The `minimize-unsafe` branch (see #29) builds on top of this to remove usage of unsafe. If you'd rather I can just PR that branch and close this one.

alicemaz:
I think we can merge in `minimize-unsafe` and proceed from there

marshallpierce:
Sounds good; closing in favor of #31.

Issue #20, opened 2017-03-31
[RFE] add support for Reader/Writer for chaining


ignatenkobrain:
My use-case is pretty simple, I have file which contains base64-encoded, zlib-compressed JSON ;)

ignatenkobrain:
I hope internally this could do something partially decoding base64 (since you can decode it by chunks), you don't need full string to be available.

marshallpierce:
Are you picturing something like a `Base64BufReader` that wraps an underlying `BufRead`?

ignatenkobrain:
yes, something like that. Unfortunately, I'm still in the middle of understanding how to chain FlateReader from flate2, so I can't come up with syntax how I would like to see it.

P.S. it's my first experience with chaining readers in Rust

ignatenkobrain:
Also I'm not sure whether it should be BufRead or just Read

ignatenkobrain:
```rust
impl EntitlementCertificate {
    pub fn from_file(file: &str) -> Self {
        let path = Path::new(file);

        let rdr = BufReader::new(File::open(path).unwrap());
        let b64iter = rdr.lines()
            .skip_while(|l| l.as_ref().unwrap() != ENTITLEMENT_DATA_HEADER)
            .skip(1)
            .take_while(|l| l.as_ref().unwrap() != ENTITLEMENT_DATA_FOOTER);
        let mut b64rdr = IterRead::new(b64iter);

        let mut data = String::new();
        b64rdr.read_to_string(&mut data).unwrap();
        let zlib_encoded = base64::decode(&data).unwrap();

        let decompressor = flate2::read::ZlibDecoder::new(&*zlib_encoded);

        serde_json::from_reader(decompressor).unwrap()
    }
}
```

that's what I've got so far in my project, as you can see, I have reading file, decompressing and deserializing working on top of Read.

brianm:
I have 80% of an implementation of this internally at work. Basically is a Read and Write implementation loosely based on the golang implementation of the streaming encoder and decoder.

I'd be happy to extract the code and contribute it if there is interest.

marshallpierce:
Sure, I'd be happy to take a look at that, if you feel like making a PR (even a rough one). I haven't forgotten about this feature; just got several things ahead of it on the to-do list. :)

BTW, thanks for JDBI!

brianm:
[...]

Issue #17, opened 2017-02-23
Display wrapper for base64 encoded value


tailhook:
Basically, I want `Base64(&[u8])` that implements `std::fmt::Display` that formats into destination buffer without intermediate memory allocations:
```
write!(file, "Sec-WebSocket-Accept: {}", Base64(value))
```

What what do you think about this?

marshallpierce:
I can definitely see the appeal. I assume heap allocation is what you want to avoid, not stack allocation?

In the current implementation, we get tremendous speed gains by writing into a buffer we control, but perhaps we can preserve the speed and do some of the buffer management for you. Perhaps if the wrapper Display struct had an array (say, 1K) and encoded its input in chunks into the array that were then emitted to the formatter, that would do the job. We would need to rework encoding a little bit to be able to output to a slice, and then converting an array full of utf-8 bytes into a Formatter, but I think it can be done.

tailhook:
Yes. Sorry that I've missed comment. I understand your point about benchmarks. I've done similar to what you desribe [here](https://github.com/fizyk20/generic-array/pull/27/files). It works.

So should I make a pull request?

marshallpierce:
Sorry for the delay; I'll take a look shortly

marshallpierce:
Do you need to support line wrapping as part of this? Supporting line wrapping when encoding in chunks that don't necessarily line up well with line length is pretty complex, so if you don't need it, I'd just as soon not bother. Labor of love only goes so far... ;)

tailhook:
No, I don't need line wrapping. Thanks for the hard work!

dtolnay:
I would have wanted to use this to serialize base64 data with Serde in https://github.com/serde-rs/json/issues/360.

marshallpierce:
#33 is merged. I'll release a new version with this and the other recently merged features once #48 gets merged or people convince me it's a bad idea.

PR #33, opened 2017-05-23
Heap-free `Display` wrapper

marshallpierce:
As per #17. @tailhook does this meet your needs? It encodes into a stack-resident buffer, then hands off the encoded data to the `Formatter` provided to `Display`.

There is a nontrivial but not overhwhelming performance cost for going through the `ChunkedEncoder` mechanism:

```
test encode_3kib                ... bench:       1,639 ns/iter (+/- 77) = 1874 MB/s
test encode_3kib_display        ... bench:       1,987 ns/iter (+/- 403) = 1546 MB/s
test encode_3mib                ... bench:   2,583,661 ns/iter (+/- 110,524) = 1217 MB/s
test encode_3mib_display        ... bench:   2,884,784 ns/iter (+/- 165,993) = 1090 MB/s
```

Also, I'm open to different names. As it currently is, I think typical usage would look like:

```
println!("Have some base64: {}", Base64Display::new(&some_bytes));
```

This opens the door to #20, at least for the encode path.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,5 +18,5 @@ safemem = "0.2.0"
 rand = "0.3.15"
 
 [profile.bench]
-# Uncomment when using `perf record`
-#debug = true
+# Useful for better disassembly when using `perf record` and `perf annotate`
+debug = true
--- a/README.md
+++ b/README.md
@@ -64,14 +64,7 @@ cargo run --example make_tables > src/tables.rs.tmp && mv src/tables.rs.tmp src/
 Profiling
 ---
 
-On Linux, you can use [perf](https://perf.wiki.kernel.org/index.php/Main_Page) for profiling. First, enable debug symbols in Cargo.toml. Don't commit this change, though, since it's usually not what you want (and costs some performance):
-
-```
-[profile.release]
-debug = true
-```
-
-Then compile the benchmarks. (Just re-run them and ^C once the benchmarks start running; all that's needed is to recompile them.)
+On Linux, you can use [perf](https://perf.wiki.kernel.org/index.php/Main_Page) for profiling. Then compile the benchmarks with `rustup nightly run cargo bench --no-run`.
 
 Run the benchmark binary with `perf` (shown here filtering to one particular benchmark, which will make the results easier to read). `perf` is only available to the root user on most systems as it fiddles with event counters in your CPU, so use `sudo`. We need to run the actual benchmark binary, hence the path into `target`. You can see the actual full path with `rustup run nightly cargo bench -v`; it will print out the commands it runs. If you use the exact path that `bench` outputs, make sure you get the one that's for the benchmarks, not the tests. You may also want to `cargo clean` so you have only one `benchmarks-` binary (they tend to accumulate).
 
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,6 +1,7 @@
 # Next
 
 - `STANDARD_NO_PAD` config
+- `Base64Display` heap-free wrapper for use in format strings, etc
 
 # 0.6.0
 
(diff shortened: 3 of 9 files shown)


tailhook:
Looks fine for my use case. Frankly, I think the implementation is too complex, I would choose simpler approach even if having more limited features. But if you need to support #20, then it might be worth it.

marshallpierce:
_Need_ is a strong word... I think it would be _interesting_ to support #20. ;) 

It definitely is somewhat on the complex side. You could convince me to ditch the line wrapping support, which would remove about half the code (and at least as much of the mental burden of understanding it). OTOH, it is at least pretty self-contained complexity. 🤔

tailhook:
The strange thing is that it's always slower than allocating version.

But if I change benchmarks:
```
diff --git a/benches/benchmarks.rs b/benches/benchmarks.rs
index 87ef5c0..358e376 100644
--- a/benches/benchmarks.rs
+++ b/benches/benchmarks.rs
@@ -231,10 +231,13 @@ fn do_encode_bench_display(b: &mut Bencher, size: usize) {
     let mut v: Vec<u8> = Vec::with_capacity(size);
     fill(&mut v);
 
+    let mut buf = String::with_capacity(size);
     b.bytes = v.len() as u64;
     b.iter(|| {
-        let e = format!("{}", display::Base64Display::new(&v));
-        test::black_box(&e);
+        use std::fmt::Write;
+        buf.truncate(0);
+        write!(&mut buf, "{}", display::Base64Display::new(&v)).unwrap();
+        test::black_box(&buf);
     });
 }
```
I get the following results:
```
test encode_3kib                ... bench:       2,335 ns/iter (+/- 57) = 1315 MB/s
test encode_3kib_display        ... bench:       2,471 ns/iter (+/- 60) = 1243 MB/s
test encode_3kib_reuse_buf      ... bench:       2,327 ns/iter (+/- 232) = 1320 MB/s
test encode_3kib_reuse_buf_mime ... bench:       2,821 ns/iter (+/- 91) = 1088 MB/s
test encode_3mib                ... bench:   3,469,686 ns/iter (+/- 77,514) = 906 MB/s
test encode_3mib_display        ... bench:   2,676,150 ns/iter (+/- 188,494) = 1175 MB/s
test encode_3mib_reuse_buf      ... bench:   2,636,884 ns/iter (+/- 63,508) = 1192 MB/s
```

Still slower on smaller chunks.

marshallpierce:
I'm moving, so I just packed up my linux systems and don't have `perf` any more, but it would be interesting to run that with `perf stat`. I bet there will be more branches on the `display` benchmarks because it has to halt every 1k (or about 30 iterations of the 32-byte encode unrolled loop) and figure out if it needs to add padding, line wraps, etc, and therefore a lower IPC.

marshallpierce:
@tailhook I think your patch has basically transformed it to be the equivalent of the `reuse_buf` variants, so it would make sense that performance would be more similar to that than to the slower variant that allocates a new `String` every time.

marshallpierce:
Having looked over this PR again after being busy for the last few weeks, it doesn't produce too many WTFs/min. However, I _do_ feel mildly queasy about the complexity of that encode logic given that nobody is begging for heap-free line-wrapped base64 output. It would be a nontrivial dent in the complexity if we just ripped out all line-wrap support and only allowed the user to specify the character set (and padding) to use. @alicemaz what do you think?

marshallpierce:
Also relevant to https://github.com/hyperium/hyper/blob/207fca63ce1073a832b5a6b08ef811d09ff4b080/src/header/common/authorization.rs#L154

marshallpierce:
@alicemaz any comments on this? Otherwise I'll go ahead and merge it

marshallpierce:
Well, I still feel good about the level of complexity of this implementation, so I'm calling it done.

PR #38, opened 2017-06-11
Add STANDARD_NO_PAD configuration.

mrijkeboer:
This adds a STANDARD_NO_PAD configuration option.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -95,6 +95,13 @@ pub static STANDARD: Config = Config {
     line_wrap: LineWrap::NoWrap,
 };
 
+pub static STANDARD_NO_PAD: Config = Config {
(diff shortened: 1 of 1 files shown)


marshallpierce:
LGTM. @alicemaz WDYT?

mrijkeboer:
Any update on this? If this is not wanted, would it be an option to make the fields of the `struct Config` public so I can make a static Config myself?

marshallpierce:
I could merge it, but I can't publish to crates.io, so it wouldn't do you much good, unfortunately.

mrijkeboer:
@marshallpierce: I would still appreciate you merging it. That way, it will be available in the next release on crates.io.

mrijkeboer:
Thanks.

PR #40, opened 2017-06-16
relax rand version

ignatenkobrain:


marshallpierce:
Just out of curiosity, what problem is this addressing?

ignatenkobrain:
@marshallpierce we are packaging this crate for Fedora and having one (same) version of rand for most of crates looks better.. I don't see any point to restrict version

marshallpierce:
I see. I think picking a single version of a library for an operating system wide installation is an anti pattern. As a `dev-depencency`, this version number shouldn't perturb any other crate, but by forcing it to be OS-wide, you're inventing this problem out of thin air. 

In this case this relaxation is harmless, so I'll merge this, but I'm not too excited about this artificial concern going forward. :/

PR #45, opened 2017-09-01
Link to the documentation on docs.rs in the README

Marwes:
At least for me I often want to navigate directly to the docs from the repository so I find it useful to have a link directly to them!


marshallpierce:
Looks good, thanks!

PR #48, opened 2017-09-22
Display wrapper API changes

marshallpierce:
While I have the opportunity to tweak this API before I release it... I just really don't like murky default choices that people will use without thinking. I figure people can make a helper function easily enough if they consistently use the same configuration.

@tailhook would this meet your needs?

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -5,7 +5,7 @@
 //! use base64::display::Base64Display;
 //!
 //! let data = vec![0x0, 0x1, 0x2, 0x3];
-//! let wrapper = Base64Display::new(&data, STANDARD).expect("STANDARD is OK");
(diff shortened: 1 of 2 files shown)


tailhook:
I'd say it's too complex for default constructor. `with_config` constructor or `config.wrap(data)` might be better. I'd also say there should be three constructors: `new`, `new_urlsafe` and full one (make a note, that urlsafe is almost always non-wrapped, because put into url). But this may stem from my use cases where base64 texts are all small and non-wrapped.

marshallpierce:
Hm, perhaps it's worth the API fluff to have some convenience methods.

tailhook:
Yeah. Looks perfect now.

PR #49, opened 2017-10-25
Add an `encode_config_slice` method to write to slices

quodlibetor:
The obvious point of this is to write to ungrowable slices and arrays.

For my case, I have some content that I can stick into the guaranteed
stack-allocated portion of an InlineString, which is pretty nice.

The idea for this came up for me because I want to serialize UUIDs to 
Base64, which has a really small encoded representation.

In case you're curious, here the commit where I implement the inline 
serialization:
https://github.com/quodlibetor/b64-ids/commit/87dd8d1de4da18d5b0b72e80900a446524ffbf65

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -218,7 +218,7 @@ pub fn encode_config<T: ?Sized + AsRef<[u8]>>(input: &T, config: Config) -> Stri
     buf
 }
 
-/// calculate the base64 encoded string size, including padding
+/// calculate the base64 encoded string size, including padding if that is in the config
 fn encoded_size(bytes_len: usize, config: &Config) -> Option<usize> {
     let rem = bytes_len % 3;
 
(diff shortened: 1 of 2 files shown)


marshallpierce:
Seems like a reasonable thing to let users do! I'll take a look when I get some time, hopefully in the next couple of days.

quodlibetor:
Awesome, thanks!

marshallpierce:
I'm working on a variant of this that has a bit more code re-use, and I'm also taking the opportunity to clean up docs, etc. 

What do you think about exposing `encoded_size()`? I think it's probably not useful in practice. If you're dynamically sizing the encode buffer, which you'd need to be doing to make use of `encoded_size()`, why wouldn't you just use `encode_config_buf()`? But maybe there's something I'm missing.

quodlibetor:
I used `encoded_size` myself to add some assertions around the tests that I was writing when doing initial discovery, so I needed it to be pub for that reason and I left it pub so that other people who wanted to use this feature could do similar tests.

The only other reason I could imagine using `encoded_size` with this features is if you need to decide whether to use a statically sized buffer or not depending on the total encoded size. I have no idea when that would be useful.

Actually, now that I've improved the error message to show the encoded size vs the available bytes (from what I originally had) I wouldn't have actually needed encoded size for my initial work.

marshallpierce:
This functionality should now be available in 0.8.0.

PR #50, opened 2017-10-31
Encode to a slice directly

marshallpierce:
Based on #49.

@quodlibetor does this work for your use case?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,51 @@
+//! # Configs
+//!
+//! There isn't just one type of Base64; that would be too simple. You need to choose a character
+//! set (standard or URL-safe), padding suffix (yes/no), and line wrap (line length, line ending).
+//! The `Config` struct encapsulates this info. There are some common configs included: `STANDARD`,
+//! `MIME`, etc. You can also make your own `Config` if needed.
+//!
+//! The functions that don't have `config` in the name (e.g. `encode()` and `decode()`) use the
+//! `STANDARD` config .
+//!
+//! # Encoding
+//!
+//! Several different encoding functions are available to you depending on your desire for
+//! convenience vs performance.
+//!
+//! | Function                | Output                       | Allocates                      |
+//! | ----------------------- | ---------------------------- | ------------------------------ |
+//! | `encode`                | Returns a new `String`       | Always                         |
+//! | `encode_config`         | Returns a new `String`       | Always                         |
+//! | `encode_config_buf`     | Appends to provided `String` | Only if `String` needs to grow |
+//! | `encode_config_slice`   | Writes to provided `&[u8]`   | Never                          |
+//!
+//! All of the encoding functions that take a `Config` will pad, line wrap, etc, as per the config.
+//!
+//! # Decoding
+//!
+//! Just as for encoding, there are different decoding functions available.
+//!
+//! | Function                | Output                        | Allocates                      |
+//! | ----------------------- | ----------------------------- | ------------------------------ |
+//! | `decode`                | Returns a new `Vec<u8>`       | Always                         |
+//! | `decode_config`         | Returns a new `Vec<u8>`       | Always                         |
+//! | `decode_config_buf`     | Appends to provided `Vec<u8>` | Only if `Vec` needs to grow    |
+//!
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -285,30 +316,85 @@ pub fn encode_config_buf<T: ?Sized + AsRef<[u8]>>(input: &T, config: Config, buf
...
+/// # Example
+/// 
+/// ```rust
+/// extern crate base64;
+/// 
+/// fn main() {
+///     let s = b"hello internet!";
+///     let mut buf = Vec::new();
+///     // make sure we'll have a slice big enough for base64 + padding
+///     buf.resize(s.len() * 4 / 3 + 4, 0);
```
quodlibetor: Oh I like this definition in the docs! But would it be worth just doing `Vec::with_capacity(...)` instead of the resize dance?

src/lib.rs
```diff
@@ -285,30 +316,85 @@ pub fn encode_config_buf<T: ?Sized + AsRef<[u8]>>(input: &T, config: Config, buf
...
+///
+///     let bytes_written = base64::encode_config_slice(s,
+///                             base64::STANDARD, &mut buf);
+///
+///     // shorten our vec down to just what was written
+///     buf.resize(bytes_written, 0);
```
quodlibetor: Is this necessary? I'm not actually sure what it accomplishes.

src/tests.rs
```diff
@@ -192,6 +192,60 @@ fn encode_into_nonempty_buffer_doesnt_clobber_existing_contents() {
...
+
+        assert_encode_sanity(std::str::from_utf8(&encoded_data[0..encoded_size]).unwrap(),
+                             &config, input_len);
+
+        assert_eq!(&encoded_data[encoded_size..], &encoded_data_original_state[encoded_size..]);
+
+        // since we know we have the correct count of line endings, it's reasonable to simply remove
+        // them without worrying about where they are
+        let encoded_no_line_endings: String = String::from_utf8(encoded_data[0..encoded_size].iter()
+            .filter(|&b| *b != 0xA_u8 && *b != 0xD_u8)
```
quodlibetor: these could be chars instead of bytes, right? https://play.rust-lang.org/?gist=5c5e84b31b8dc90037fb9e03e0ae6b99

src/lib.rs
marshallpierce: `with_capacity` specifies the internal capacity but doesn't extend the length of the vec. I need the elements to really be there so that when I take a mutable slice it includes all those bytes (set to 0 here).

src/lib.rs
marshallpierce: We made the `Vec` conservatively sized originally, so if we didn't truncate it back down, we'd have some stray 0 bytes that would prevent the subsequent `decode` from working.

src/tests.rs
marshallpierce: Ah, good call. I think it's slightly more readable to cast from `char` to `u8`, which also seems to work well. Good thing those are all single-byte chars in UTF8.

src/lib.rs
quodlibetor: Oh right I always seem to forget (when working with b64 in particular, for some reason) that you can't have uninitialized data in a slice.

quodlibetor:
This looks great to me!

quodlibetor:
Also I've confirmed that my tests work without modification when using this branch, which is expected.

marshallpierce:
ping @alicemaz in case you have input

src/lib.rs
marshallpierce: Yeah, it's a bit awkward -- I wish there was a type-safe way to say "give me some data that I'm about to overwrite". Fortunately modern CPUs do a great job of splatting 0 into memory.

shaleh:
Is there a reason why we can encode to a slice but need to decode to a vector? I am bridging C and Rust so Vectors don't really work well. If the answer is the code hasn't been written I might be able to make a PR happen.

marshallpierce:
@shaleh It's simply that nobody asked for it yet. :) I've opened https://github.com/alicemaz/rust-base64/issues/52.

shaleh:
Thank you @marshallpierce

PR #53, opened 2017-12-05
Decode slice

### LINKED ISSUE #52: Decode to a slice
For when a Vec is inconvenient.

marshallpierce:
Fixes #52.

This turned out to be quite a bit more work than I expected, but the results are pretty good I think.

First, a bcmp vs master:

```
 name                    control.bcmp ns/iter    branch.bcmp ns/iter     diff ns/iter   diff % 
 decode_100b             96 (1041 MB/s)          106 (943 MB/s)                    10   10.42% 
 decode_100b_reuse_buf   73 (1369 MB/s)          80 (1250 MB/s)                     7    9.59% 
 decode_10mib            7,852,829 (1335 MB/s)   7,098,399 (1477 MB/s)       -754,430   -9.61% 
 decode_10mib_reuse_buf  5,992,922 (1749 MB/s)   5,293,776 (1980 MB/s)       -699,146  -11.67% 
 decode_30mib            24,089,655 (1305 MB/s)  22,006,699 (1429 MB/s)    -2,082,956   -8.65% 
 decode_30mib_reuse_buf  18,534,301 (1697 MB/s)  16,336,266 (1925 MB/s)    -2,198,035  -11.86% 
 decode_3b               42 (95 MB/s)            79 (50 MB/s)                      37   88.10% 
 decode_3b_reuse_buf     24 (166 MB/s)           29 (137 MB/s)                      5   20.83% 
 decode_3kib             1,676 (1832 MB/s)       1,533 (2003 MB/s)               -143   -8.53% 
 decode_3kib_reuse_buf   1,673 (1836 MB/s)       1,500 (2048 MB/s)               -173  -10.34% 
 decode_3mib             2,225,701 (1413 MB/s)   2,045,099 (1538 MB/s)       -180,602   -8.11% 
 decode_3mib_reuse_buf   1,721,010 (1827 MB/s)   1,542,305 (2039 MB/s)       -178,705  -10.38% 
 decode_500b             307 (1628 MB/s)         291 (1718 MB/s)                  -16   -5.21% 
 decode_500b_reuse_buf   287 (1742 MB/s)         267 (1872 MB/s)                  -20   -6.97% 
 decode_50b              66 (787 MB/s)           75 (693 MB/s)                      9   13.64% 
 decode_50b_reuse_buf    46 (1130 MB/s)          54 (962 MB/s)                      8   17.39% 
```

Some regressions on extremely short inputs, but worth it for the speedups elsewhere. More good news on how well `decode_slice` does even compared to `master`'s performance on short inputs:

```
test decode_100b                ... bench:         106 ns/iter (+/- 6) = 943 MB/s
test decode_100b_reuse_buf      ... bench:          80 ns/iter (+/- 6) = 1250 MB/s
test decode_10mib               ... bench:   7,098,399 ns/iter (+/- 407,102) = 1477 MB/s
test decode_10mib_reuse_buf     ... bench:   5,293,776 ns/iter (+/- 858,381) = 1980 MB/s
test decode_30mib               ... bench:  22,006,699 ns/iter (+/- 733,699) = 1429 MB/s
test decode_30mib_reuse_buf     ... bench:  16,336,266 ns/iter (+/- 526,990) = 1925 MB/s
test decode_30mib_slice         ... bench:  14,875,003 ns/iter (+/- 547,959) = 2114 MB/s
test decode_3b                  ... bench:          79 ns/iter (+/- 2) = 50 MB/s
test decode_3b_reuse_buf        ... bench:          29 ns/iter (+/- 0) = 137 MB/s
test decode_3b_slice            ... bench:          18 ns/iter (+/- 0) = 222 MB/s
test decode_3kib                ... bench:       1,533 ns/iter (+/- 61) = 2003 MB/s
test decode_3kib_reuse_buf      ... bench:       1,500 ns/iter (+/- 47) = 2048 MB/s
test decode_3kib_slice          ... bench:       1,467 ns/iter (+/- 44) = 2094 MB/s
test decode_3mib                ... bench:   2,045,099 ns/iter (+/- 116,055) = 1538 MB/s
test decode_3mib_reuse_buf      ... bench:   1,542,305 ns/iter (+/- 49,346) = 2039 MB/s
test decode_3mib_slice          ... bench:   1,487,613 ns/iter (+/- 38,308) = 2114 MB/s
test decode_500b                ... bench:         291 ns/iter (+/- 9) = 1718 MB/s
test decode_500b_reuse_buf      ... bench:         267 ns/iter (+/- 9) = 1872 MB/s
test decode_50b                 ... bench:          75 ns/iter (+/- 1) = 693 MB/s
test decode_50b_reuse_buf       ... bench:          54 ns/iter (+/- 4) = 962 MB/s
test decode_50b_slice           ... bench:          43 ns/iter (+/- 1) = 1209 MB/s
```

Worst case on 3 byte decodes is a lot slower than `master`, but best case is a lot faster. On non-tiny inputs, 2GiB/s is finally attainable on my E5-1650v3.

At a high level, the logic changes were to rearrange the decode loops to never write garbage bytes past the end of what would eventually be overwritten with more decoded bytes. We used to get away with this because we could just `truncate()` the vec, but it seemed unacceptable to require the output buffer to be oversized by 2 bytes in some cases: users should be able to use an output buffer that's exactly the size of the decoded data.

### DIFF
--- a/benches/benchmarks.rs
+++ b/benches/benchmarks.rs
@@ -4,7 +4,7 @@ extern crate base64;
 extern crate rand;
 extern crate test;
 
-use base64::{decode, decode_config_buf, decode_config_slice, encode, encode_config_buf, Config, MIME, STANDARD};
+use base64::{decode, decode_config_buf, decode_config_slice, encode, encode_config_buf, encode_config_slice, Config, MIME, STANDARD};
 use base64::display;
 
 use test::Bencher;
@@ -21,6 +21,11 @@ fn encode_3b_reuse_buf(b: &mut Bencher) {
 }
 
 #[bench]
+fn encode_3b_slice(b: &mut Bencher) {
+    do_encode_bench_slice(b, 3, STANDARD)
+}
+
+#[bench]
 fn encode_50b(b: &mut Bencher) {
     do_encode_bench(b, 50)
 }
@@ -33,6 +38,11 @@ fn encode_50b_display(b: &mut Bencher) {
 #[bench]
 fn encode_50b_reuse_buf(b: &mut Bencher) {
     do_encode_bench_reuse_buf(b, 50, STANDARD)
+}
+
+#[bench]
+fn encode_50b_slice(b: &mut Bencher) {
+    do_encode_bench_slice(b, 50, STANDARD)
 }
 
 #[bench]
@@ -76,6 +86,11 @@ fn encode_3kib_reuse_buf(b: &mut Bencher) {
 }
 
 #[bench]
+fn encode_3kib_slice(b: &mut Bencher) {
+    do_encode_bench_slice(b, 3 * 1024, STANDARD)
+}
+
+#[bench]
 fn encode_3kib_reuse_buf_mime(b: &mut Bencher) {
     do_encode_bench_reuse_buf(b, 3 * 1024, MIME)
 }
@@ -96,6 +111,11 @@ fn encode_3mib_reuse_buf(b: &mut Bencher) {
 }
 
 #[bench]
+fn encode_3mib_slice(b: &mut Bencher) {
+    do_encode_bench_slice(b, 3 * 1024 * 1024, STANDARD)
+}
+
+#[bench]
 fn encode_10mib(b: &mut Bencher) {
     do_encode_bench(b, 10 * 1024 * 1024)
 }
@@ -113,6 +133,11 @@ fn encode_30mib(b: &mut Bencher) {
 #[bench]
 fn encode_30mib_reuse_buf(b: &mut Bencher) {
     do_encode_bench_reuse_buf(b, 30 * 1024 * 1024, STANDARD)
+}
+
+#[bench]
+fn encode_30mib_slice(b: &mut Bencher) {
+    do_encode_bench_slice(b, 30 * 1024 * 1024, STANDARD)
 }
 
 #[bench]
@@ -292,6 +317,20 @@ fn do_encode_bench_reuse_buf(b: &mut Bencher, size: usize, config: Config) {
     b.iter(|| {
         encode_config_buf(&v, config, &mut buf);
         buf.clear();
+    });
+}
+
+fn do_encode_bench_slice(b: &mut Bencher, size: usize, config: Config) {
+    let mut v: Vec<u8> = Vec::with_capacity(size);
+    fill(&mut v);
+
+    let mut buf = Vec::new();
+
+    b.bytes = v.len() as u64;
+    // conservative estimate of encoded size
+    buf.resize(size * 2, 0);
+    b.iter(|| {
+        encode_config_slice(&v, config, &mut buf);
     });
 }
 
--- a/benches/benchmarks.rs
+++ b/benches/benchmarks.rs
@@ -4,7 +4,7 @@ extern crate base64;
 extern crate rand;
 extern crate test;
 
-use base64::{decode, decode_config_buf, encode, encode_config_buf, Config, MIME, STANDARD};
+use base64::{decode, decode_config_buf, decode_config_slice, encode, encode_config_buf, Config, MIME, STANDARD};
 use base64::display;
 
 use test::Bencher;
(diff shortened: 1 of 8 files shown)


marshallpierce:
@shaleh does this address your need?

shaleh:
@marshallpierce thank you for doing this work. I have been nibbling at my own attempt.

I will grab your branch and test it for matching our needs.

shaleh:
I have not run it yet but this looks good.

However..... yup, me moving the goal post.

This crate panicking is also an issue for me. I want to handle failures in application code without needing to unwind a panic. Ideally all functions in this crate return a Result or Option, at least as a choice. `decode_blah_safe` maybe.

marshallpierce:
I can definitely understand that desire. However, I can't do that without using checked writes everywhere, and that would give up a lot of the speed because it relies on bulk slice to slice operations under the hood.

Can you give a bit more detail about your use case? In practice, it will never fail if you give it enough room, and you can use a conservative estimate of how much room will be required. Also, I could expose `encoded_size` if you need that on the encode side -- I've held off on that because I'd rather not increase the public API if there isn't a need for it.

shaleh:
[Emacs ported to Rust](https://github.com/Wilfred/remacs)

We obviously do not want to have the editor crash because we got the math wrong :-) Wrapping every crate we use in unwind logic is annoying to say the least.

You have a panic! on line 301 of decode.rs. That is not a memory size issue right?

shaleh:
Performance is important but not our biggest concern. We are not churning out tons of encoded/decoded data. The biggest use of base64 in Emacs is Tramp. Emacs allows the user to edit remote files by sending the buffers back and forth over ssh base64 encoded.

marshallpierce:
Looks like it's on line 395 to me, but I take it you're referring to the one near the end of `decode_helper` (god, what an awful function name)? That one is an impossible condition -- it should not be possible to hit any number beyond what is special cased because the number is at most 8 and 1 and 5 have already been checked at the start of decoding.

I'm sympathetic to your goal of not crashing (and in other code that's not giving up as much performance I sprinkle around `Result` with abandon), but realistically would you not simply `unwrap()` or `expect()` such a `Result` from a base64 decode anyway? Either it works, always, or it doesn't. To me, this particular situation (an undersized buffer where there is always a conservative estimate that will work) is a programmer error that should be fixed like any other bug, not a transient error like i/o issues that might need a retry.

If your goal really is that nothing should ever panic, then I take it that you are not using `some_slice[foo]` anywhere, or `Vec` for that matter, which will panic if `usize` overflows as it grows...? There may well be a good answer here, but I'm not seeing it at the moment.

FWIW, http://joeduffyblog.com/2016/02/07/the-error-model/ has been formative on my opinions on error handling.

shaleh:
We do not unwrap anything without knowing it is safe beforehand. All returns are tested and transformed into some form of handled failure. The code does have aborts/asserts/panics but that is inherited from GNU Emacs and not our preference.

Crashing may be ok for internet services. Spawning another is easy and no one notices. Dying in the middle of editing someone's code or dissertation is not the same thing. Do that too often and no one will use the application.

shaleh:
If you have panics where you believe they can not be reached, consider using `unreachable!` to document them to the reader and the compiler.

marshallpierce:
Good point re `unreachable!`. I had waffled on that a bit, but if it seems clearer to someone !me then it's probably a better choice.

My point is not just about `unwrap()` on a particular thing, it's that there's an awful lot of stdlib that panics in certain cases, so you're already exposed to a lot of that. If you hope to avoid *all* situations that can panic if you write a bug, you'll need to start from basically zero. There are also non-panic situations that can kill you, like allocating a few dozen GiB in a Vec and having the oom killer kill your process.

shaleh:
Yeah. I am learning that Rust is not as safe as they promise sadly. Hopefully as the language reaches more people that will change. Until then we all do what we can to make sane error handling be the norm.

shaleh:
I will try out the code as soon as I can. Hopefully tonight, might be a day or 3 though.

marshallpierce:
No, Rust is pretty darn safe. I think you're confusing "can't crash" with "can't corrupt memory" -- the former is MUCH less of a problem than the latter. :)

shaleh:
This is working well so far.

marshallpierce:
Great! I'll let it percolate for another day or two in case @alicemaz wants to review it, and so that I can look at it with fresh eyes before merging.

marshallpierce:
This is released in 0.9.0.

shaleh:
Thank you! We will provide feedback if anything comes up.

PR #54, opened 2018-01-01
bump rand to 0.4

ignatenkobrain:


marshallpierce:
I don't have a problem with this change; I'm just curious what prompted it. Care to elaborate?

ignatenkobrain:
In Fedora we try to keep only latest versions of each crate to avoid maintenance burden.

PR #58, opened 2018-04-17
Add the `crypt(3)` character set

tjamaan:
Add the `crypt(3)` character set. This character set is used by the `crypt(3)` C function commonly used in password hashing (e.g. `/etc/passwd` in Linux, hashed passwords stored in databases, etc...).


marshallpierce:
Thanks for the contribution. It looks like you left out the change to `make_tables.rs`, though.

tjamaan:
Please let me know if I missed anything else 🙂

tjamaan:
PS: I wanted to use inclusive ranges but they are not yet stable.

marshallpierce:
Thanks. I'll take a detailed look as soon as I can, but it might be a week or so.

marshallpierce:
Looks good, thanks. I'll release this as 0.9.1.

PR #62, opened 2018-06-14
no_std support

astro:
Could be useful for embedded systems.

Thank you for this project!

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,14 @@ cargo +nightly fuzz run roundtrip_mime -- -max_len=10240
 cargo +nightly fuzz run roundtrip_random_config -- -max_len=10240
 ```
 
+no_std support
+---
+
+This crate has a feature flag `std` that is enabled by default.
+
+To use in a *no_std* environment, you can disable default
+features. You must setup an allocator because this API makes use of
+the dynamically-sized `Vec` and `String` types.
 
 License
 ---
--- a/README.md
+++ b/README.md
@@ -102,9 +102,13 @@ no_std support
 
 This crate has a feature flag `std` that is enabled by default.
 
-To use in a *no_std* environment, you can disable default
-features. You must setup an allocator because this API makes use of
-the dynamically-sized `Vec` and `String` types.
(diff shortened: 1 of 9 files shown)


marshallpierce:
Thanks for this; I'll take a look. Unfortunately it will probably take me several weeks at least to get to this. I'm moving right now, so my spare time goes towards putting things in cardboard boxes.

linkmauve:
This looks correct to me, I also tested it on no_std (but with alloc available) and it works as expected, thanks!

README.md
```diff
@@ -97,6 +97,18 @@ cargo +nightly fuzz run roundtrip_mime -- -max_len=10240
...
+
+This crate has a feature flag `std` that is enabled by default.
+
+To use in a *no_std* environment, you can disable default features. To
+use the full API, enable feature `alloc` and setup an allocator.
+
+In environments without an allocator, the API provides only the
+`encode_config_slice()` and `decode_config_slice()` functions (which
+do not use the dynamically-sized `Vec` type) and the
+`strip_whitespace` config must be disabled.
```
marshallpierce: This unpleasant interplay of features is yet more evidence to me that line wrapping (and its dual, whitespace stripping) are a mistake to include, at least at this level. Maybe a `base64::ext` should contain helper functions to do that stuff after the fact. This would also simplify #56.

marshallpierce:
Line wrapping is gone, and I'm going to remove whitespace stripping as well, which should make this tidier to implement.

CryZe:
What's the status of this?

marshallpierce:
I'd like to have this feature shipped; it's just waiting on my free time. 

Now that the `alloc` crate is stable in 1.36, we could use that, and I'd be happy to bump the minimum version to 1.36 to ship that feature. (This would also answer the version question in #110 ).

If you feel like picking up the pieces from this long since bitrotted PR, that'd be great, but if not, it'll be several months likely until I can dedicate the time to do it myself.

marshallpierce:
This is now done via #111.

PR #63, opened 2018-06-16
bump safemem to 0.3

ignatenkobrain:


marshallpierce:
Thanks!

stevenroose:
Why was the major version of safemem updated without bumping the major version of base64??

See https://github.com/marshallpierce/rust-base64/issues/112.

PR #65, opened 2018-07-02
Change the suggested version in the README

linkmauve:
The README was currently suggesting to install the 0.6 branch, which is now quite outdated.

It may be useful to add a note in your release process to not forget to update this example.


marshallpierce:
Thanks! I think I shall eliminate it entirely and add a link to the crate instead.

Issue #64, opened 2018-06-25
`cargo doc` breaks on 1.28.0-beta.2


ErichDonGubler:
I'm going to bet that this is an upstream problem -- pretty sure you ARE actually using `safemem`. But `cargo +beta doc` doesn't think so in the newly-cut beta. :(

<details>
<summary>Console log output</summary>

```
~/…/bug-reports/rust-base64 master$ rustc --version
rustc 1.28.0-beta.2 (5981061e5 2018-06-22)

~/…/bug-reports/rust-base64 master$ cargo doc
    Updating registry `https://github.com/rust-lang/crates.io-index`
 Downloading safemem v0.3.0
    Checking byteorder v1.2.3
    Checking safemem v0.3.0
 Documenting safemem v0.3.0
 Documenting byteorder v1.2.3
 Documenting base64 v0.9.2 (file:///C:/Users/egubler/workspace/personal/bug-reports/rust-base64)
error: unused extern crate
 --> src\line_wrap.rs:1:1
  |
1 | extern crate safemem;
  | ^^^^^^^^^^^^^^^^^^^^^ help: remove it
  |
note: lint level defined here
 --> src\lib.rs:59:57
  |
59|     missing_docs, trivial_casts, trivial_numeric_casts, unused_extern_crates, unused_import_braces,
  |                                                         ^^^^^^^^^^^^^^^^^^^^

error: Compilation failed, aborting rustdoc

error: Could not document `base64`.

Caused by:
  process didn't exit successfully: `rustdoc --crate-name base64 src\lib.rs -o C:\Users\egubler\workspace\personal\bug-reports\rust-base64\target\doc -L dependency=C:\Users\egubler\workspace\personal\bug-reports\rust-base64\target\debug\deps --extern byteorder=C:\Users\egubler\workspace\personal\bug-reports\rust-base64\target\debug\deps\libbyteorder-5d6212fcf45e1323.rmeta --extern safemem=C:\Users\egubler\workspace\personal\bug-reports\rust-base64\target\debug\deps\libsafemem-8a9b675cbece16a6.rmeta` (exit code: 101)
```
</details>

marshallpierce:
Thanks, I'll take a look.

marshallpierce:
All fixed on the latest nightly and beta.

robinst:
Would you mind if I created a PR to add a build using Travis?

marshallpierce:
Not at all. See #68.

PR #66, opened 2018-07-05
Fix `cargo doc` breaks on latest rust 1.28.0-nightly

zzeroo:
This addresses #64

I'm unsure if this is right.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "base64"
-version = "0.9.2"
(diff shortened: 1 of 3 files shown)


marshallpierce:
Thanks for the contribution. I don't love disabling the `warnings` check; I'll see if I can find a way to avoid that.

marshallpierce:
Since the rust tools folks took care of #64, we don't need this any more, and I can stop musing on how to avoid disabling warnings.

PR #74, opened 2018-10-06
Streaming encoder

marshallpierce:
Supersedes #56. 

cc @brianm @nox


ignatenkobrain:
Wait, no DecodeWriter?

marshallpierce:
Well, this PR is entitled "Streaming encoder"... ;) One thing at a time. Most of this logic can eventually be shared with a `Write` that encodes, and there will also eventually be `Read` impls as well.

marshallpierce:
@nox does this suit your use case?

nox:
Yes!

brianm:
Yea!

nox:
This may be broken.

https://github.com/servo/servo/pull/22147#issuecomment-437341675

marshallpierce:
Looking into it

nox:
@marshallpierce found out that the problem is probably in the png crate, which uses `io::Write::write` without checking how many bytes were actually written.

I've opened https://github.com/PistonDevelopers/image-png/pull/94. There is no issue with the base64 crate AFAICT.

PR #77, opened 2018-10-18
Run cargo-tarpaulin on nightly Rust

robinst:
See README of tarpaulin: https://github.com/xd009642/tarpaulin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ matrix:
   include:
     - rust: 1.23.0
     - rust: stable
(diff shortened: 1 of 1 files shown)


robinst:
The report is a bit verbose. We can disable it if you want by adding a config file: https://docs.codecov.io/docs/pull-request-comments#section-disable-comment

marshallpierce:
Seems fine for now; we can always revisit if it gets annoying. Thanks!

Issue #34, opened 2017-05-24
Ideas for removing the last `unsafe`


briansmith:
The work to remove the other uses of `unsafe` was awesome. However, I do think it is worthwhile to try to remove all of the uses of `unsafe`, of which there is currently only one. I think a lot of people are like me and think that `unsafe` should only be used as a last resort. I think there are alternatives that are workable and even better in terms of usability for certain circumstances than what we have now.

/cc @marshallpierce @alicemaz

Here are three ideas for removing the last use of `unsafe`:

1. Just do it, on the assumption the difference in `encode_config_buf()` performance isn't that critical, because very few applications need high-performance base64 *encoding* of large data.

2. Change the signature to use move semantics:

Old:
```rust
pub fn encode_config_buf<T: ?Sized + AsRef<[u8]>>(input: &T, config: Config, buf: &mut String) {
}
```
New:
```
pub fn encode_config_buf<T: ?Sized + AsRef<[u8]>>(input: &T, config: Config, buf: String) -> String {
    let buf_bytes = buf.into_bytes();
    ....
    String::from_utf8(buf_bytes)
}
```

3. Change the signature to use `&mut Vec` instead of `&mut String`. A caller that has a `String` would then be able to decide whether they want to use `as_mut_vec()` for performance reasons, or whether they want to use move semantics (like suggestion 2 above), or whether they want to start with a `Vec`. This also has the advantage that if you have a `Vec<u8>` already, e.g. because you're writing base64 data into some kind of binary or otherwise non-UTF-8 array, you don't require any extra allocations.

marshallpierce:
I like the third approach -- I'd already been mulling the merits of allowing users to provide a `Vec<u8>` for convenience reasons, and the lack of unsafe in that code path would be a nice benefit. I wonder if maybe we should even expose the ability to write to a slice?

Also note that #33 involves one other unsafe (doing the same sort of thing as the existing one).

marshallpierce:
#33 no longer uses unsafe (but users have the option to reimplement a bit of it if they need the performance). 

Creating a variant that encodes into a `Vec<u8>` is easy enough. Supporting the existing `encode_config_buf<T: ?Sized + AsRef<[u8]>>(input: &T, config: Config, buf: &mut String)` is pretty tricky without unsafe, though. 

- We could allocate an extra `Vec<u8>`, encode into that, then append it to the string. This is easy, but incurs allocation overhead and of course a memory copy.
[...]

PR #79, opened 2018-10-26
Deoptimise encoding into string and deny(unsafe_code)

ctz:
This PR is a bit speculative. See issue #34 for more discussion.

It takes the first option mentioned in https://github.com/alicemaz/rust-base64/issues/34#issuecomment-313542184 -- as such it'll add an extra allocation and copy when encoding into a String.

This PR also adds `unsafe_code` to the crate-wide lint denies.

### DIFF
--- a/src/encode.rs
+++ b/src/encode.rs
@@ -1,5 +1,6 @@
 use byteorder::{BigEndian, ByteOrder};
 use {line_wrap, line_wrap_parameters, Config, LineWrap, STANDARD};
+use std::str;
 
 ///Encode arbitrary octets as base64.
 ///Returns a String.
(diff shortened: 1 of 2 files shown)


ctz:
> Thanks, I'll take a look. Is there a particular reason for this change (aside from generic "unsafe makes the world a worse place")?

There's not much more to it than that :)

The background was: I was running cargo-geiger on my downstream crate that depends on base64, and base64 is the only crate where unsafe is used but isn't fundamental to the purpose of the crate.

marshallpierce:
Fair enough. I'm guessing rustls? I want rustls to be unsafe-free. :)

I'm working on the line wrap removal this weekend. Hopefully I'll get that done quickly, at which point I'll benchmark this approach vs ChunkedEncoder and pick whichever is faster.

marshallpierce:
This was pretty painful, almost halving encode speed in most cases. #80 is a bit more complex but much less costly.

PR #80, opened 2018-10-28
Encode without unsafe

marshallpierce:


marshallpierce:
This way has much less performance cost than #79. Also addresses #34.

ctz:
Thanks! :)

Issue #76, opened 2018-10-17
Reject improperly encoded base64 trailing symbols


marshallpierce:
As per https://github.com/alicemaz/rust-base64/issues/75#issuecomment-430621532.

marshallpierce:
Fixed.

PR #82, opened 2018-10-28
Detect trailing symbols that encode unused bits and report an error.

marshallpierce:
Addresses #76.

@AljoschaMeyer WDYT?

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -6,6 +6,7 @@
 - Add a streaming encoder `Write` impl to transparently base64 as you write.
 - Remove the remaining `unsafe` code.
 - Remove whitespace stripping to simplify `no_std` support. No out of the box configs use it, and it's trivial to do yourself if needed: `filter(|b| !b" \n\t\r\x0b\x0c".contains(b)`.
+- Detect invalid trailing symbols when decoding and return an error rather than silently ignoring them.
 
 # 0.9.3
 
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -744,7 +744,7 @@ mod tests {
 
     #[test]
     fn detect_invalid_last_symbol_every_possible_three_symbols() {
-        let mut base64_to_bytes = std::collections::HashMap::new();
+        let mut base64_to_bytes = ::std::collections::HashMap::new();
 
         let mut bytes = [0_u8; 2];
         for b1 in 0_u16..256 {
@@ -753,7 +753,7 @@ mod tests {
                 bytes[1] = b2 as u8;
                 let mut b64 = vec![0_u8; 4];
                 assert_eq!(4, ::encode_config_slice(&bytes, STANDARD, &mut b64[..]));
-                let mut v = std::vec::Vec::with_capacity(2);
+                let mut v = ::std::vec::Vec::with_capacity(2);
                 v.extend_from_slice(&bytes[..]);
 
                 assert!(base64_to_bytes.insert(b64, v).is_none());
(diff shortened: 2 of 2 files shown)


AljoschaMeyer:
LGTM, to stick with the four-letter abbreviations.

You could also add an additional roundtrip fuzzer. The current ones encode the random input and then decode. With this merged, you could also do a fuzzer that tries to decode the random input, do nothing if parsing fails, otherwise reencode and check for equality with the input. That's how I discovered the issue in the first place.

I'd love to see a release of this. For your amusement, here is my temporary workaround code:

```rust
// XXX temporary until https://github.com/alicemaz/rust-base64/issues/76 is published
pub(crate) fn is_canonical(data: &[u8]) -> bool {
    if data.len() < 2 {
        return false;
    }
    if data[data.len() - 2] == b"="[0] {
        if data.len() < 3 {
            return false;
        }
        let b = data[data.len() - 3];
        return b == b"A"[0] || b == b"Q"[0] || b == b"g"[0] || b == b"w"[0];
    } else {
        if data[data.len() - 1] == b"="[0] {
            let b = data[data.len() - 2];
            return b == b"A"[0] || b == b"E"[0] || b == b"I"[0] || b == b"M"[0] ||
            b == b"Q"[0] || b == b"U"[0] || b == b"Y"[0] || b == b"c"[0] ||
            b == b"g"[0] || b == b"k"[0] || b == b"o"[0] || b == b"s"[0] ||
            b == b"w"[0] || b == b"0"[0] || b == b"4"[0] || b == b"8"[0];
        } else {
            true
        }
    }
}
```

marshallpierce:
Dear lord that workaround is horrendous!

I haven't found a way to make fuzzers actually useful in practice. I doubt Travis is going to be very happy if I burn up infinite CPU time running them. Are there other infrastructure choices out there for that sort of thing?

AljoschaMeyer:
I haven't combined fuzzers and CI yet. I suppose you could write a script that stops them after a timeout? For a crate like this one, a few minutes of fuzzing are probably enough to explore the complete state space.

marshallpierce:
@nox interesting, so it does happen in the wild then. Would making it configurable whether or not that error is emitted work in your use case?

nox:
Yes.

https://html.spec.whatwg.org/multipage/webappapis.html#dom-btoa

Some of the failures:

```
  FAIL atob("ab") - The string contains invalid characters.
  FAIL atob("ab==") - The string contains invalid characters.
  FAIL atob("ab\t\n\f\r =\t\n\f\r =\t\n\f\r ") - The string contains invalid characters.
  FAIL atob("A/") - The string contains invalid characters.
  FAIL atob("AA/") - The string contains invalid characters.
  FAIL atob("YR") - The string contains invalid characters.
  FAIL atob(12) - The string contains invalid characters.
  FAIL atob(NaN) - The string contains invalid characters.
```

Note that `12` and `NaN` result in the strings `"12"` and `"NaN"` being parsed.

nox:
I guess what I need is a proper implementation of https://infra.spec.whatwg.org/#forgiving-base64-decode

marshallpierce:
I see. I shouldn't be surprised that the web contains such horrors, I suppose. ;) It looks to me like `forgiving_base64_decode` should be what you get when that error isn't emitted.

nox:
@marshallpierce Unfortunately, the `Config` struct being what it is, adding a flag there means a new breaking bump. :(

nox:
Or we could piggy-back a Builder pattern on top of `Config`, adding a `forgiving(&mut self, bool)` method to it.

marshallpierce:
I'm OK with that. I take an aggressive view towards API evolution... if the current API is bad, I'd rather change it than make users suffer with the wrong API for an extended time. I'm going to cut a release shortly with #92. If an 0.11 with the above config change (hmm, will think about builder... it solves for *adding* config things but not *removing*, but then again there isn't much left to remove) will solve your problem that's fine with me.

nox:
Ah, if you cut a release shortly, I think `Config::new` should be killed then, and two methods `pad` and `forgiving` should be added to `Config`. That way new additions can be done in a backwards-compatible way (and most people just use `base64::STANDARD` and friends anyway).

marshallpierce:
Duly noted. https://infra.spec.whatwg.org/#forgiving-base64-decode

PR #84, opened 2018-11-20
Implement SSE and AVX2 encoding

ggriffiniii:
This adds an implementation of B64 encoding using SSE and AVX2 intrinsics. This can make a substantial difference especially on AVX2 enabled processors. Detailed benchmark results on both an ivybridge (SSE only) and skylake (AVX2) are availabe at https://ggriffiniii.github.io/rust-base64/reports/

The implementation is based off the description here: http://www.alfredklomp.com/programming/sse-base64/

Some of the highlights from the benchmarks:

encode_slice on just 3 bytes of input slightly degrades:
<pre>
skylake:   309.32 MiB/s => 292.89 MiB/s
ivybridge: 277.94 MiB/s => 258.61 MiB/s
</pre>
I tried to minimize this regression as much as possible, but I'm afraid this is the best I could do. In exchange for the small regression on small inputs the performance on large inputs increases substantially.

encode_slice on 100 bytes of input:
<pre>
skylake:   1.0509 GiB/s => 2.1306 GiB/s
ivybridge: 1.3279 GiB/s => 1.4122 GiB/s
</pre>
encode_slice on 3KiB of input:
<pre>
skylake:   1.9132 GiB/s => 4.7891 GiB/s
ivybridge: 1.5108 GiB/s => 1.7158 GiB/s
</pre>
In addition to the SSE/AVX2 this pull request also includes commits that apply rustfmt, fix clippy warnings, and move from the standard benchmark harness to criterion. Let me know if you want any of those commits dropped from the pull request.


ggriffiniii:
Sorry, I realized when I rebased this pull request I reintroduced a previously fixed bug. I'll open a new pull request that fixes it. Sorry for the noise.

PR #85, opened 2018-11-20
Implement SSE and AVX2 encoding

ggriffiniii:
This adds an implementation of B64 encoding using SSE and AVX2 intrinsics. This can make a substantial difference especially on AVX2 enabled processors. Detailed benchmark results on both an ivybridge (SSE only) and skylake (AVX2) are availabe at https://ggriffiniii.github.io/rust-base64/reports/

The implementation is based off the description here: http://www.alfredklomp.com/programming/sse-base64/

Some of the highlights from the benchmarks:

encode_slice on just 3 bytes of input slightly degrades:
<pre>
skylake:   309.32 MiB/s => 292.89 MiB/s
ivybridge: 277.94 MiB/s => 258.61 MiB/s
</pre>
I tried to minimize this regression as much as possible, but I'm afraid this is the best I could do. In exchange for the small regression on small inputs the performance on large inputs increases substantially.

encode_slice on 100 bytes of input:
<pre>
skylake:   1.0509 GiB/s => 2.1306 GiB/s
ivybridge: 1.3279 GiB/s => 1.4122 GiB/s
</pre>
encode_slice on 3KiB of input:
<pre>
skylake:   1.9132 GiB/s => 4.7891 GiB/s
ivybridge: 1.5108 GiB/s => 1.7158 GiB/s
</pre>
In addition to the SSE/AVX2 this pull request also includes commits that apply rustfmt, fix clippy warnings, and move from the standard benchmark harness to criterion. Let me know if you want any of those commits dropped from the pull request.


marshallpierce:
Thanks, SIMD has been on my todo list for a long time. Why have merely fast base64 when it could be completely unnecessarily fast?

Can you split out rustfmt/clippy/criterion into its own PR to make it easier to see which are the SIMD changes?

I won't have time for a detailed review of the SIMD work for a while, but just as food for thought -- a design I explored in https://bitbucket.org/marshallpierce/stream-vbyte-rust/src that worked out pretty well was to have a trait that represented the core encoding/decoding logic with implementations for scalar and all the relevant flavors of SIMD. That helped keep it (arguably?) more readable and maintainable. I'd be inclined to follow that to make conceptual room for NEON implementations, etc, but I haven't taken a close look at what you've got here yet.

ggriffiniii:
Yes, I can split out the rustfmt/clippy/criterion changes into a separate PR. Would you like those as 3 PR's or just 1?

Regarding the SIMD implementation. I had considered a trait and it would be a straightforward change if you prefer. In this case I didn't think it was worth it since it's just a single function (bulk_encode_slice) and putting the function within a module seemed like it accomplished much of the same goal as a trait would, but I can see how a trait could potentially evolve more gracefully when adding decode, other architectures, etc.

I noticed that the continuous build failed trying to build against rust 1.23. Is that the minimum rust version you're targeting? If so, this may not be possible yet. This change would require 1.27 or later if the simd functionality is enabled.

ggriffiniii:
Also I'm sure you realize, but GitHub will show diffs for individual commits if you click on them. I'm fine splitting the change into multiple pull requests still. Just felt obligated to point that out.

marshallpierce:
Just one PR for the precursor commits is fine -- that one I should be able to get merged quickly. Might as well have that much done and easily available for other work.

Let's leave the code structure as is for now. It'll be easy enough to switch to traits if need be.

SIMD is definitely worth bumping the minimum rust version for.

marshallpierce:
Also worth looking into from my stash o' bookmarks, if you hadn't seen them already:

https://github.com/lemire/fastbase64
https://arxiv.org/pdf/1704.00605.pdf
http://richardstartin.uk/base64-encoding/ for more context on the above
http://0x80.pl/notesen/2016-04-03-avx512-base64.html for AVX512

ggriffiniii:
I'm going to close this pull request as it's probably suffered bitrot over the months. If you're still interested in implementing this there is an improved version in my base64 library `radix64` https://github.com/ggriffiniii/radix64 

Feel free to take any bits you like.

PR #88, opened 2018-11-26
Use Quickcheck for property based testing.

ggriffiniii:
I'm not tied to this change at all and feel free to reject anything you don't like.

Prior to this commit a variety of property based tests existed and used
a manual approach, consisting of filling vectors with random data and
choosing random configs using the rand crate. This moves to using
quickcheck to handle a lot of the grunt work that was previously done
manually. In addition to the simpler code quickcheck also provides
"shrinking" of input data to reduce test cases when possible. This can
make troubleshooting test failures much simpler.

In the process of refactoring the tests it became clear that there
exists a lot of redundancy. I took the liberty of removing a variety of
test cases that entirely overlapped with an already existing case. I
believe the end result is net gain in terms of readability and
maintenance of the test cases without any loss of code coverage.


marshallpierce:
Thanks; will take a look. I'm sure there are duplicate tests. They've accreted somewhat haphazardly and could use some pruning.

marshallpierce:
I incorporated this in #108; thanks for laying the groundwork. I didn't remove some of the tests, though, as they were originally written to test certain thresholds (when we go from a slow loop to a fast loop, etc) and I'm not completely sure that they're covered elsewhere. Since they're not causing a maintenance headache (yet), I left them in.

PR #90, opened 2018-12-09
fix(encoder-writer): use write_all instead of write

dignifiedquire:
If the inner writer does not write everything, the missing data is never written
using write_all, ensures all data is written to the inner writer

### DIFF
--- a/src/write/encoder.rs
+++ b/src/write/encoder.rs
@@ -119,7 +119,7 @@ impl<'a, W: Write> EncoderWriter<'a, W> {
                 &mut self.output[..],
             );
             self.panicked = true;
-            let _ = self.w.write(&self.output[..encoded_len])?;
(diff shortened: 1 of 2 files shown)


dignifiedquire:
Maybe I am missing something, but this the way that `BufWriter` solves this problem in the std lib effectively: https://doc.rust-lang.org/src/std/io/buffered.rs.html#473, so it seems there is some different interpretation of this or I am missing sth.

marshallpierce:
Given that the above issue is unlikely to be resolved quickly, I think I'll proceed with an implementation that returns `Ok(0)` and caution people to be careful about `write_all`. You can always write your own straightforward `write_all` clone in a couple of lines that doesn't do the nasty behavior with `Ok(0)`...

marshallpierce:
Closing this in favor of #92, which honors the contract of `write` by breaking `write_all` in some cases... :'(

PR #92, opened 2018-12-23
Streaming encoder: keep track of encoded bytes that weren't written.

marshallpierce:
They'll be retried on subsequent writes, but this plays poorly with
write_all. See https://github.com/rust-lang/rust/issues/56889.

Hat tip to #90.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -2,6 +2,10 @@
 
 - Minimum rust version 1.27.2
 
+# 0.10.1
+
+- Fix bug in streaming encoding ([#90](https://github.com/alicemaz/rust-base64/pull/90)): if the underlying writer didn't write all the bytes given to it, the remaining bytes would not be retried later. See the docs on `EncoderWriter::write`.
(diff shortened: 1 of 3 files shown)


marshallpierce:
@dignifiedquire any thoughts on this implementation? Does this address your issue?

dignifiedquire:
My current code relies heavily on `write_all`, so I can't easily test it, but it does look like it achieves what it needs to.

marshallpierce:
Would it be feasible in your codebase to have a `write_all_safe` defined for some `WriteExtension` trait that doesn't choke on `Ok(0)` and use that instead? I'm wondering if that's maybe something I could suggest in the docs for how to get past the `write_all` limitations while the rust core team figures out what a better path forward would be.

dignifiedquire:
Yes I probably need to do sth like that. I'll see if I find some time to do this later this week

marshallpierce:
You mean, why does `write_all` behave the way it does? It doesn't make sense to me either.

PR #93, opened 2019-01-19
Use sha2 crate

newpavlov:
Using `ring` for a single SHA-256 function is certainly an overkill.

BTW running `cargo check` on `fuzz` crate produces the following error:
```
error[E0061]: this function takes 2 parameters but 3 parameters were supplied
 --> fuzzers/roundtrip_no_pad.rs:6:18
  |
6 |     let config = base64::Config::new(base64::CharacterSet::Standard, false, false);
  |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 2 parameters

```

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -11,7 +11,7 @@ cargo-fuzz = true
 [dependencies]
 rand = "0.6.1"
 rand_pcg = "0.1.1"
-ring = "0.13.5"
+sha2 = "0.7"
 [dependencies.base64]
 path = ".."
 [dependencies.libfuzzer-sys]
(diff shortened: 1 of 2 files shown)


marshallpierce:
Thanks for the contribution. I'm trying to find an elegant way to have travis build (but not spend a lot of time running) fuzzers to make sure that they don't break. However, I don't really buy the "overkill" argument for ring vs rust-crypto. ring is actively maintained and about twice as fast at hashing. What is the incentive to switch away? The code is about the same in this crate either way.

marshallpierce:
Those statements are true. I just don't see them as relevant for a fuzzer. Can you describe a workflow that is affected by those issues?

newpavlov:
Well, if pulling the whole `ring` just for one hash function is alright with you, you can close this PR. :/

As for speed: 1) IIUC `data` is too small for it to matter 2) `sha2` can use assembly instead of pure-Rust code, and we plan to use OpenSSL assembly soon, so performance will be equalized in future.

marshallpierce:
OK, thanks for the comments. I don't view dependency size as relevant for fuzzing.

PR #103, opened 2019-05-12
Add no_std support using alloc if std feature is disabled

akhilles:
Tried to minimize code impact by not adding pure `no_std`. `no_std` with `alloc` should still be very useful.


marshallpierce:
Thanks, I'll take a look soon.

PR #106, opened 2019-06-22
Add a streamer reader

nwalfield:
This MR adds a streaming reader similar to the streaming writer that is in write/encoder.rs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -74,6 +74,7 @@ mod chunked_encoder;
 pub mod display;
 mod tables;
 pub mod write;
+pub mod read;
 
 mod encode;
 pub use encode::{encode, encode_config, encode_config_buf, encode_config_slice};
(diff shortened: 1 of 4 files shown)


nwalfield:
The CI failure happens using the nightly version of rustc.  Specifically, the error message is:

```
error: trait objects without an explicit `dyn` are deprecated
  --> src/decode.rs:58:32
   |
58 |     fn cause(&self) -> Option<&error::Error> {
   |                                ^^^^^^^^^^^^ help: use `dyn`: `dyn error::Error`
   |
note: lint level defined here
  --> src/lib.rs:61:5
   |
61 |     warnings,
   |     ^^^^^^^^
   = note: #[deny(bare_trait_objects)] implied by #[deny(warnings)]
error: aborting due to previous error
```

I think that this is a preexisting problem and not due to my patch.

marshallpierce:
Thanks! I'll take a look as soon as I can.

marshallpierce:
Working on this in #122 -- unfortunately we need more complex state management in case the delegate writer does short reads and we end up with partial base64 chunks, etc. But, I'm poking at it, and it shouldn't be too awful to rearrange.

PR #111, opened 2019-08-12
Implement Support for no_std

CryZe:
This adds the features `std` and `alloc` to the crate. The `std` feature is activated by default. This would bump the minimum supported version to 1.34.0, but only due to the byteorder removal (which apparently is also happening in #110).

Supersedes #62


CryZe:
alloc is only used when the std feature is deactivated and the alloc feature is. So 1.34.0 can still be the minimum version (but lower is also possible, we only use it for the byte reading stuff).

For CI we can compile to thumbv6m-none-eabi, which doesn't have std. I'll look into making it compile it that. However we can't run the tests then.

I'll put a section into the README.

CryZe:
This should be good to go now.

marshallpierce:
Great, thank you!

PR #113, opened 2019-08-15
Make compatible with Rust 1.19.0

stevenroose:
Apart from trivial styling things like not using `use crate::` or prefixing `#[rustfmt::skip]`, there was only one real feature of the newer Rust versions that was being used: `slice::rotate_left`. I added an implementation based on a C++ implementation that I found and it passes all the tests.

It amazes me that just for this one method, this crate requires Rust v1.31.0 instead of v1.19.0.

I didn't make the tests compatible with the same Rust version. Somehow crossbeam is pulled in as a test dependency and it's not compatible with Rust v1.19.0 no v1.24.0.

I'm not very fluent with Travis config files, so it's possible my edits need some tweaking.

I'm running the benchmarks for comparison, but I don't expect anything outrageous.


CryZe:
Rust's rotate seems to be much more optimized though: https://doc.rust-lang.org/beta/src/core/slice/rotate.rs.html#64

stevenroose:
I ran the benchmarks and I didn't see any compromise tbh. It's a method that's not used in the main API functions. I think the different might really be negligible.

Sadly I don't think it's possible to have `#[cfg()]` conditional about compiler versions. Is there? Or a conditional on the existence of rotate?

CryZe:
You can do something like this: https://github.com/seanmonstar/unicase/blob/master/build.rs

stevenroose:
@marshallpierce I don't understand. Did you look at the diff? There is literally one single std function that makes base64 incompatible with older versions, all the rest is just formality stuff. You're not losing any functionality or performance.

stevenroose:
This is the main base64 crate on crates.io. Are you just personally deciding that users of older compilers have to go and write their own base64 crate because you don't want to write `extern crate` lines in your `lib.rs`?

stevenroose:
> Rust 1.19 came out over 2 years ago. Thanks for the effort, but supporting old compilers is not high on the priority list.

I just picked 1.19 because it's the oldest version that has some relevance for some people and that could easily be made compatible without more changes from f.e. 1.24.0 which is the last version of Debian.

The README states

> The minimum required Rust version is 1.31.0.

v1.31.0 is just 9 months old.

stevenroose:
> Rust 1.19 came out over 2 years ago. Thanks for the effort, but supporting old compilers is not high on the priority list.

I just picked 1.19 because it's the oldest version that has some relevance for some people and that could easily be made compatible without more changes from f.e. 1.24.0 which is the last version of Debian.

The README states

> The minimum required Rust version is 1.31.0.

v1.31.0 is not even 9 months old.

I don't understand. Did you look at the diff? There is literally one single std function that makes base64 incompatible with older versions, all the rest is just formality stuff. You're not losing any functionality or performance.

This is the main base64 crate on crates.io. Are you just personally deciding that users of older compilers (mainline Debian users, mrustc users) have to go and write their own base64 crate because you don't want to accept `extern crate` lines in your `lib.rs`?

marshallpierce:
Ahh yes, being hassled by petulant users who haven't yet realized that _the universe does not revolve around you and your needs_. This is why I contribute to open source, right?

I do, in fact, have goals for this project that apparently are not 100% aligned with your monastic "I will only use what Debian stable uses, despite their historical unsuitability as an oracle of quality for much of anything". Goals like SIMD, and no_std, and so forth. You are welcome to use an equally ancient version of this, and other, libraries with your ancient compiler. In the meantime, progress marches on. You do realize that companies charge a boatload for enterprise support contracts to support old versions of things, right? They do that because it's time consuming and annoying, and most people don't need the result anyway. If you had approached the problem as "here's this annoying thing we'd like to see, and we're happy to pay for the inconvenience and time it will take", that's one thing. But that's not how it went, was it.

But ultimately, I'm tired of reading your blithely self-centered comments. I will give you the following choice:
- You can leave this project alone for a few months. Go on a vacation, hike up something tall and contemplate the infinite, learn how to not be so abrasive to people who donate their time to projects, try sous vide cooking, communicate via stating goals and why they're relevant to you rather than presenting solutions that address only your needs and insisting that they are the only reasonable ones, or if you're a night owl, perhaps astrophotography. Then, with a renewed sense of collaboration and a more cheerful attitude, you are welcome to return.
- Or you can continue to be a pest and I'll just block you until I remember to un-block you. Given that I tend to spend my time doing productive stuff and not staring at my GitHub settings, it might be a while.

Continued silence means you have chosen door number 1 - a better option all around. Anything else means door number 2.

stevenroose:
I went on vacation. After being scolded like that, no renewed sense of collaboration, however. I'd like to very briefly address some of your remarks and then you won't see me here again.

> petulant users who haven't yet realized that the universe does not revolve around you and your needs.

I have no petulant needs whatsoever. I don't personally use an old version of Rust, but I work on libraries that has users that do. So that's why I care about an ecosystem where software can last longer than a few months.

> You do realize that companies charge a boatload for enterprise support contracts to support old versions of things, right? 

You do realize that in high-security environments, companies require entire software stacks to be audited and approved, right? You can't expect those companies to pay for a full audit of the Rust stack every month.

> If you had approached the problem as "here's this annoying thing we'd like to see, and we're happy to pay for the inconvenience and time it will take", that's one thing. But that's not how it went, was it.

I literally implemented the change... What can you expect more? Me to pay you for the time it takes to merge it? Send me a tip link, I'll send you a buck or two, sure.

Anyway, I think whether or not it makes sense to have support for older Rust versions is not the point here. How can a crate as trivial as a base64 encoder *need* features that Rust v1.30.0 and less don't have. My main point is that the changes required to support Rust v1.19.0 are minimal, almost non-existent and furthermore they are not even breaking any of the existing API or functionality. So there is literally no downside other than having to split up your `use` statements instead of combining then in a fancy way.

So, I'll leave it at this, feel free to still use my PR to include it in this crate. In the meantime, I published `base64-compat` which includes the PR: https://crates.io/crates/base64-compat

marshallpierce:
Cool, door #2 it is. Thanks for confirming your complete insensitivity to any goals that don't match your own.

PR #116, opened 2019-09-07
Advertise that unsafe code is forbidden

Shnatsel:
Having zero unsafe code and forbidding it is a desirable property for Rust code. This commit adds a badge to advertise this property. This helps users make a more informed choice when picking a crate for base64 encoding/decoding.

In this PR clicking the badge leads to the Rust Secure Code WG coordination repo for removing unnecessary unsafe code from the Rust ecosystem. I'm fine with making the link lead somewhere else or making the badge non-clickable if you prefer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 [base64](https://crates.io/crates/base64)
 ===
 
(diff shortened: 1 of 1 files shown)


marshallpierce:
LGTM, thank you!

PR #122, opened 2019-11-20
nwalfield's streaming reader

marshallpierce:
Continuing from #106.


marshallpierce:
@nwalfield take a look when you get a chance

PR #124, opened 2019-12-01
Eliminated a call to encoded_size function

imor:
The encode_config function called encode_config_slice which in turn
called encode_with_padding. Both encode_config and encode_config_slice
called encoded_size resulting in encoded_size being called twice. This
commit changes the encode_config function to directly call the
encode_with_padding function, saving one call to encoded_size.


marshallpierce:
Thanks, will get to it when I can.

marshallpierce:
Thanks for the PR; merged via #128.

PR #126, opened 2020-03-05
Relaxed ?Sized requirement for decode and encode functions

frol:
It seems that the ?Sized requirement is redundant.

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -81,7 +81,7 @@ impl error::Error for DecodeError {
 ///}
 ///```
 #[cfg(any(feature = "alloc", feature = "std", test))]
-pub fn decode<T: ?Sized + AsRef<[u8]>>(input: &T) -> Result<Vec<u8>, DecodeError> {
+pub fn decode<T: AsRef<[u8]>>(input: T) -> Result<Vec<u8>, DecodeError> {
     decode_config(input, STANDARD)
 }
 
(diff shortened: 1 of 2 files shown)


frol:
NOTE: CI failure on nightly is not related to the PR changes. All tests just pass.

marshallpierce:
Isn't `?Sized` _removing_ a constraint? https://doc.rust-lang.org/std/marker/trait.Sized.html

frol:
It removes the constraint from the `T` (which is `AsRef<[u8]>`), but it enforces the reference (which is Sized), and this does not make as lot of sense with `AsRef` trait. Consider the following example:

```
fn qq1<T: AsRef<[u8]> + ?Sized>(_a: &T) {}

fn qq2<T: AsRef<[u8]>>(_a: T) {}

fn main() {
    let a = vec![1];
    qq1(&a);
    //qq1(a);  // <-- This does not compile
    qq2(&a);
    qq2(a);
}
```

I think `AsRef<[u8]>` is the same as `&(AsRef<[u8]> + ?Sized)`, but the latter has a restriction to pass by an explicit reference.

marshallpierce:
Gotcha. I'll fix Travis.

marshallpierce:
merge in `master`, and then this should be ready to go.

frol:
@marshallpierce Done.

marshallpierce:
Thanks!

PR #132, opened 2020-05-05
Add a bcrypt alphabet

Zireael-N:


marshallpierce:
Thanks. One wonders what fever dream caused them to use a different character set to begin with...

marshallpierce:
Released in 0.12.1.

Zireael-N:
> One wonders what fever dream caused them to use a different character set to begin with...

I once had a pleasure of working with a base64 implementation that used N most significant bits where commonly N least significant bits were used and vice versa, that was an experience I don't think I'll ever forget.

marshallpierce:
dear  god D:

PR #133, opened 2020-05-07
Replace "cfg(test)" with "cfg(doctest)" for readme testing

GuillaumeGomez:
Rustdoc now passes "doctest" when running in test mode.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -76,11 +76,11 @@ extern crate alloc;
 #[cfg(any(feature = "std", test))]
 extern crate std as alloc;
 
-#[cfg(test)]
+#[cfg(doctest)]
 #[macro_use]
 extern crate doc_comment;
 
-#[cfg(test)]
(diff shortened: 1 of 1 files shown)


marshallpierce:
Thanks!

ctz:
This has broken a downstream crate's MSRV :( Until the tooling for managing rustc versions is better in cargo, could I request that changes that move your MSRV are included only in major version changes?

GuillaumeGomez:
A solution would be to put this behind a `cfg_attr`.

marshallpierce:
@GuillaumeGomez can you explain your suggestion in a bit more detail? My cfg-fu is weak and I'm not sure how to solve this with cfg_attr.

GuillaumeGomez:
Like passing a feature only on CI. For example "ci-test", which would give in the code:

```rust
#[cfg_attr(feature = "ci-test", doctest)]
```

Or something along the line.

PR #136, opened 2020-06-25
Remove running tests in the readme

### LINKED ISSUE #135: Minimum supported Rust version was bumped in 0.12.2
The changes in https://github.com/marshallpierce/rust-base64/commit/93fe059a1d3d5234be0945f89221e36054895801 cause 0.12.2 to be incompatible with 1.39.0, and rustls (as one user of base64) fixed base64 at 0.12.1. This is a bit unfortunate -- would it be okay to roll back the change for a while? Since base64 is such a low-level crate, it seems like being careful about minimum supported Rust version would be good.

marshallpierce:
It is too much of a pain to support stable (which supports `cfg(doctest)`), 1.39.0 (which has it, but as an experimental feature), and 1.34.0 (which doesn't have it at all).

Fixes #135.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ sudo: required
 matrix:
   include:
     - rust: 1.34.0
+    # cfg(doctest) is experimental in 1.39 but ignored with 1.34.0, and that snuck in when 1.39.0 wasn't tested
+    - rust: 1.39.0
     - rust: stable
     - rust: beta
     - rust: nightly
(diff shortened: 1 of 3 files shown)


djc:
FYI, you could also consider using [skeptic](https://crates.io/crates/skeptic). err-derive (which I maintain) has an example of using it only when a particular feature is enabled, so that it doesn't cause issues for downstream users.

https://gitlab.com/torkleyy/err-derive/-/blob/master/build.rs

marshallpierce:
Thanks, I'll take a look when I get a chance.

PR #138, opened 2020-07-28
Make methods of `Config` const functions

AlephAlpha:
Make the following functions `const fn` s:

* `Config::new`
* `Config::pad`
* `Config::decode_allow_trailing_bits`


marshallpierce:
Thanks. I'll look into what's going on with Travis.

marshallpierce:
#140 should have appeased Travis.

PR #142, opened 2020-09-24
Add EncoderStringWriter

marshallpierce:


kevincox:
A better way to do this is probably to leave it as a ref and you can call [`String.as_mut_vec()`](https://doc.rust-lang.org/std/string/struct.String.html#method.as_mut_vec) to make the `EncoderStringWriter`. This way you don't need to own the value and don't break backwards compatibility.

src/write/encoder_string_writer.rs
```diff
@@ -0,0 +1,107 @@
...
+    ///
+    /// Returns the base64-encoded form of the accumulated written data.
+    ///
+    /// # Errors
+    ///
+    /// The first error that is not of `ErrorKind::Interrupted` will be returned.
+    pub fn finish(&mut self) -> io::Result<String> {
+        let buf = self.encoder.finish()?;
+
+        let str = String::from_utf8(buf).expect("Base64 should always be valid UTF-8");
```
kevincox, line 55: Call me risky but I would use `String::from_utf8_unchecked()` and add a `debug_assert!()` that is is valid. There is no reason to check it at runtime when we know that we just generated valid utf8.

marshallpierce: I just benchmarked to see, and it's only about a 5% performance difference until inputs get very large, at which point I assume people aren't typically buffering 10MiB of data on the heap, but if they are, they'll probably still be ok with 1.7GiB/s encoding instead of 1.9 :)

marshallpierce:
Surprisingly (to me), it looks like the new API for `EncoderWriter` is actually compatible, and thus a superset -- note that tests using `&mut delegate` didn't have to change.

kevincox:
Hmm, that being compatible does make sense now that you mention it. I'll continue the review then.

src/write/encoder_string_writer.rs
kevincox: Also note that microbenchmarks aren't always the best. In this case unicode parsing (at least the subset done by `&str`) can take up valuable instruction cache space. However your stance is completely valid so I will resolve this discussion. It can also always be changed later.

src/write/encoder.rs
```diff
@@ -89,38 +86,58 @@ impl<'a, W: Write> fmt::Debug for EncoderWriter<'a, W> {
...
+    /// Returns the writer that this was constructed around.
     ///
     /// # Errors
     ///
-    /// The first error that is not of [`ErrorKind::Interrupted`] will be returned.
-    pub fn finish(&mut self) -> Result<()> {
-        if self.finished {
-            return Ok(());
+    /// The first error that is not of `ErrorKind::Interrupted` will be returned.
+    pub fn finish(&mut self) -> Result<W> {
```
kevincox, line 119: I think the more idiomatic action would be to call this `into_inner()` and make it take `self`. This also removes the need for `Option<W>` in the body. However that would be a less subtle API break and would be quite likely to cause issues.

Also note that this is an API break because of the return type. People who were using it in a position where the value was being "used" (probably after a `?`) will get an error now because `W != ()`.

src/write/encoder_string_writer.rs
marshallpierce: Point well taken. This library used to use a _lot_ of unsafe and I did get it all out eventually, so I'm hesitant to lose my `forbid(unsafe_code)`. It does pain me to lose the performance, but eventually in some bizarro universe where I have time to do it, SIMD will more than make up for the UTF-8 cost.

src/write/encoder_string_writer.rs
```diff
@@ -0,0 +1,107 @@
...
+///
+/// Calling `write()` (or related methods) or `finish()` after `finish()` has completed without
+/// error is invalid and will panic.
+///
+/// # Performance
+///
+/// B64-encoded data is buffered in the heap since the point is to collect it in a String.
+pub struct EncoderStringWriter {
+    encoder: EncoderWriter<Vec<u8>>,
+}
```
kevincox, line 34: It would be nice if this could write to an existing string. This would mean making it similar to the `EncoderWriter` struct taking a reference to a possibly non-empty `String`.

src/write/encoder_string_writer.rs
```diff
@@ -0,0 +1,107 @@
...
+    /// associated padding.
+    ///
+    /// Once this succeeds, no further writes or calls to this method are allowed.
+    ///
+    /// Returns the base64-encoded form of the accumulated written data.
+    ///
+    /// # Errors
+    ///
+    /// The first error that is not of `ErrorKind::Interrupted` will be returned.
+    pub fn finish(&mut self) -> io::Result<String> {
```
kevincox, line 52: Since there is no existing users to satisfy I would make this consume self. This removes the possibility of truing to use it after `finish`.

```suggestion
    pub fn finish(mut self) -> io::Result<String> {
```

src/write/encoder_string_writer.rs
```diff
@@ -0,0 +1,107 @@
...
+        let buf = self.encoder.finish()?;
+
+        let str = String::from_utf8(buf).expect("Base64 should always be valid UTF-8");
+        Ok(str)
+    }
+}
+
+impl<'a> Write for EncoderStringWriter {
```
kevincox, line 60: This shouldn't be implemented as it allows a user to write invalid utf8 to the backing string which will `panic` upon calling `finish`.

src/write/encoder.rs
kevincox: Another option here is to keep `finish` with the previous signature and add an `into_inner()` method with the new signature. It wouldn't allow you to get rid of the `Option` (and associated complexity) but it might make the change fully backwards compatible. You can then could mark `finish` as deprecated and one day remove it (and the `Option`).

marshallpierce: Ah you are right! I had actually done this first and then undid it (see comment below) but finish() is _not_ retryable because it has to have an internal loop to keep writing, which means that interrupted errors are already ignored. So, problem solved there.

Re naming: to me I wouldn't expect into_inner() to perform I/O, but maybe that precedent is well laid elsewhere?

src/write/encoder_string_writer.rs
marshallpierce: Good point; no reason why not.

src/write/encoder_string_writer.rs
marshallpierce: It writes via the underlying `EncoderWriter`, so this will all end up UTF8'd.

marshallpierce (commented):
Thanks for the careful look!

src/write/encoder.rs
kevincox: There is precedent for IO in std: https://doc.rust-lang.org/std/io/struct.BufWriter.html#method.into_inner

src/write/encoder_string_writer.rs
kevincox: Oh derp, I misunderstood.

src/write/encoder.rs
marshallpierce: I'm okay with the pretty minor compatibility load of adding a `let _ = `. I would bet that all but the most careful people aren't calling finish() to begin with, and are just letting `Drop` take care of it. Hopefully the careful people won't mind...

marshallpierce: Doh -- turns out I still need the `Option`, as I cannot move the `W` out of `EncoderWriter` since it implements `Drop`.

Isn't it breaking API, regardless of return type, to make `finish` take `self`?

kevincox: It's up to you. Personally I would prefer deprecating this to make it take self by value but you are probably right that few if any users will break and that the fix is very easy. At least for my use I am relying on `Drop` so I have no horse in the race.

kevincox: > Doh -- turns out I still need the `Option`, as I cannot move the `W` out of `EncoderWriter` since it implements `Drop`.

Ah right. I forget about this rust "bug".

> Isn't it breaking API, regardless of return type, to make `finish` take `self`?

It definitely is. You would want to make a different method.

src/write/encoder_string_writer.rs
marshallpierce: :thinking: Should this just return `String`? I don't see how it can fail:
- Encoding base64 has no error cases
- The underlying I/O target is a `Vec<u8>`, which shouldn't ever fail to write (unless you're out of memory, in which case the game is already over)

kevincox: Yeah, I'm pretty sure you can safely unwrap it. The `impl Write for Vec` can't fail: https://doc.rust-lang.org/src/std/io/impls.rs.html#367-399

src/write/encoder.rs
marshallpierce: I think I was wrong with my first comment here: finish() is retryable because my policy of retrying on interrupt is just that: _my_ policy (based on write_all's logic of nebulous validity). Another user might decide that other `ErrorKind`s are retryable in their use case, so it would be inappropriate to consume `self` if `finish()` returned `Err` that they might want to retry on.

kevincox: I don't think there is any sort of general guarantee that `write_all` (or `write`) don't do partial writes (or full writes but return an error anyways). I guess since the user knows what `Write` implementation they are using they can be smart about the retries if they make assumptions about how this type works.

So you are right, in the general case retries are not safe, but depending on what the user chooses to assume about this crate and the `Write` impl they can meaningfully retry.

marshallpierce: With `into_inner` out of reach since it can't consume `self`, got any ideas for an expressive name for `finish but also return the delegate writer`? If there's something that's not too odious, you've convinced me that that's worth the trouble to stay compatible.

kevincox: I can't think of a good name. I think if it can't be made to consume `self` than it is less attractive to make a new method.

src/write/encoder_string_writer.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// associated padding.
+    ///
+    /// Once this succeeds, no further writes or calls to this method are allowed.
+    ///
+    /// Returns the base64-encoded form of the accumulated written data.
+    pub fn into_inner(mut self) -> String {
+        let buf = self.encoder.finish()
+            .expect("Writing to a Vec<u8> should never fail");
+
+        String::from_utf8(buf).expect("Base64 should always be valid UTF-8")
```
joshtriplett, line 57: It would be nice to accumulate directly into a `String`, to avoid first writing to a `Vec<u8>` and then copying into a `String`.

kevincox: See https://github.com/marshallpierce/rust-base64/pull/142#discussion_r494625657 for the previous discussion.

src/write/encoder_string_writer.rs
kevincox: I don't think this is resolved yet unless you forgot to push some code.

src/write/encoder_string_writer.rs
kevincox: To be clear converting `Vec -> String` does not require any copying (it is a free conversion) however the function used first checks that the bytes are valid utf8 to upload the `String`'s invariant.

joshtriplett: Ah, I misread and didn't catch that this was taking ownership of the Vec and not copying the memory. Nevermind, then.

I'm not concerned about paying the cost for the UTF-8 validity checking; it'd be nice to avoid that, but not worth adding unsafe code for.

src/write/encoder_string_writer.rs
marshallpierce: I added `from()` below (open to suggestions on a better name) -- you give it a string, it gives it back to you via into_inner().

src/write/encoder_string_writer.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+impl EncoderStringWriter {
+    /// Create a new EncoderStringWriter that will encode with the provided config.
+    pub fn new(config: Config) -> EncoderStringWriter {
+        EncoderStringWriter::from(String::new(), config)
+    }
+
+    /// Create a new EncoderStringWriter that will append to the provided string.
+    pub fn from(s: String, config: Config) -> EncoderStringWriter {
```
kevincox, line 43: Taking by value is very restrictive. It would be much more continent if it took a `&mut String`. For example if you had a function that would append to a `&mut String` you couldn't call this. However you can always convert a `String` to a `&mut String`.

marshallpierce: I don't disagree it would be more convenient, but how would it be structured since I wouldn't be able own the underlying `Vec<u8>`? Perhaps with a new abstraction like `Utf8OnlyWriter` that could convert the incoming base64 bytes into a `str` to be `push_str()`'d onto the String?

kevincox: It does require `unsafe` but you can just instantiate a `EcoderWriter<&mut Vec<u8>>` by calling `as_mut_vec()`.

marshallpierce: :( I _really_ want to stay unsafe free. I'll see what I can do.

kevincox: I don't think it is possible then unless you make this class independent of `EncodeWriter`.

Actually you could probably make this the base class and push `char`s onto the string. Then `as_vec()` is safe and free. I don't know the performance of pushing `char` but I if you do `s.push(byte as char)` the compiler needs to notice that `byte < 128`. Actually doing it directly into a `String` is probably "more correct" since you are building a string and know that it will be valid.

src/write/encoder_string_writer.rs
```diff
@@ -0,0 +1,173 @@
...
+/// ```
+///
+/// Or, append to an existing String:
+///
+/// ```
+/// use std::io::Write;
+///
+/// let mut buf = String::from("base64: ");
+///
+/// let mut enc = base64::write::EncoderStringWriter::from(&mut buf, base64::STANDARD);
```
marshallpierce, line 33: @kevincox have your cake and eat it too, via yet another layer of abstraction

src/write/encoder_string_writer.rs
```diff
@@ -0,0 +1,173 @@
...
+        self.encoder.flush()
+    }
+}
+
+/// An abstraction around infallible writes of `str`s.
+///
+/// Typically, this will just be String.
+pub trait StrWrite {
+    /// The write must succeed, and must write the entire `buf`.
+    fn write(&mut self, buf: &str);
```
marshallpierce, line 97: maybe this shouldn't be called write since it doesn't error? Or maybe it should return `Result<()>` to make it possible to have this wrap an actual I/O type, though I'm not sure how it would be possible to implement it safely since the semantics require that the write is atomic.

src/write/encoder_string_writer.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    fn flush(&mut self) -> io::Result<()> {
+        self.encoder.flush()
+    }
+}
+
+/// An abstraction around infallible writes of `str`s.
+///
+/// Typically, this will just be String.
+pub trait StrWrite {
```
marshallpierce, line 95: better names?

src/write/encoder_string_writer.rs
marshallpierce: see `StrWriter`

src/write/encoder_string_writer.rs
kevincox: Haha, that works!

src/write/encoder_string_writer.rs
kevincox: It reminds me of https://doc.rust-lang.org/std/iter/trait.Extend.html. I don't think you can use `Extend` directly though because I don't think it is implemented for `&mut String`.

You also might be able to use [`AsMut<String>`](https://doc.rust-lang.org/std/convert/trait.AsMut.html)? I think that would let you use `String` or `&String`?

src/write/encoder_string_writer.rs
kevincox: You could call it `push` to match Vec and String. But `write` sounds fine to me.

kevincox (approved):

src/write/encoder_string_writer.rs
marshallpierce: > You also might be able to use [`AsMut<String>`](https://doc.rust-lang.org/std/convert/trait.AsMut.html)? I think that would let you use `String` or `&String`?

Do you mean as a way to avoid implementing `StrWrite` for `&mut StrWrite`? I think I want that no matter what though so that it's not _only_ available for `String` and `&mut String`.

kevincox: No, I think it may avoid the need for the trait overall. Because you can use AsMut to convert `&mut String` and `&mut mut String` to `&mut String` which will just allow you to call `String.push_str` directly.

kevincox: Hmm, I'm having trouble getting the this to work.

kevincox: Ah, I had the wrong trait. You can use `BorrowMut`.

https://rust.godbolt.org/z/bPd9z9

marshallpierce: Good to know. I think I'll keep the StrWrite though so it's a step closer to being no_std friendly -- currently all that's needed to finish that effort would be to introduce an abstraction other than `Write` for `EncoderWriter`, AFAICT.

kevincox:
`BorrowMut` is in core if that is your only concern.

https://doc.rust-lang.org/core/borrow/trait.BorrowMut.html

Of course you aren't writing to `String`s in `no_std` anyways IIUC so I
think you would need to disable this anyways.

marshallpierce:
There wouldn't be String certainly but someone could have a non-String `StrWrite`, I would imagine.

I'm still not totally happy with `StrWrite` as a name because it's mirroring `Write`, and yet offers rather different semantics (no errors). What about `StrSink` or `StrConsumer`?

kevincox:
Ah, I see what you mean. In that case it makes sense but I think you would have to drop the `&str` argument. 

However if you wanted to make something that is useful for core you are probably better off making an API that works with bytes again. 

🤷‍♀️

marshallpierce:
Is `&str` not available without std? I thought it would be, since it doesn't mandate allocation, etc.

kevincox:
I missed it originally because it isn't in `core` but `str` is actually a primitive type so it is available without `std` so you are right.

PR #144, opened 2020-09-26
Give better error messages when decoding data with trailing newlines

untitaker:
See #105


marshallpierce:
Thanks; I'll follow up with docs etc.

PR #146, opened 2020-10-17
Added lifetime parameters to all the writing systems

sapient-cogbag:
I don't think this is a breaking change as far as I am aware.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "base64"
-version = "0.13.0"
+version = "0.13.1"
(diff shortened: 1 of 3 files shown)


marshallpierce:
Thanks; I will look into this as soon as I can.

marshallpierce:
Can you help me understand the use case for this? Since [Write is implemented for &mut Write](https://doc.rust-lang.org/src/std/io/impls.rs.html#52-82), why do we need to add lifetimes to make it reference-friendly?

marshallpierce:
Stale

Issue #137, opened 2020-07-08
Feature request: expose encoded_size() function


Veetaha:
Hi, it would be awesome if this crate exposed this function:
https://github.com/marshallpierce/rust-base64/blob/ad28eadc9886b6803a0e61f96e5f3fecbd9646f5/src/encode.rs#L284-L305

Having this function we would be able to know how many bytes to preallocate before encoding.
This is useful when we put the base64 sequence in the middle of the string, such that we could preallocate the memory for surrounding contents + base64 sequence.

My use case
I need to base64-encode protobuf message into a double-quoted string.
I know how many bytes protobuf message will take, but I don't know how many it will take in base64 form beforehand.

marshallpierce:
That seems reasonable, though note that for your use case it may be ok to use a conservative estimate like `((byte len / 3) + 1) * 4`.

If the conservative estimate doesn't work for you, though, I'd be happy to take a PR to expose it.

Veetaha:
Yup, I can definitely go with that, I'll submit a PR when I'd definitely require this feature, will leave the issue open for anyone else who might be interested to leave feedback

marshallpierce:
Done in https://github.com/marshallpierce/rust-base64/pull/167.

PR #149, opened 2020-11-06
expose encoded_size function

GallagherCommaJack:
solves #137

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -85,7 +85,7 @@ mod tables;
 pub mod write;
 
 mod encode;
-pub use crate::encode::encode_config_slice;
+pub use crate::encode::{encode_config_slice, encoded_size};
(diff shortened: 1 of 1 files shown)


marshallpierce:
Just curious -- what is your use case that the conservative 4/3 estimate doesn't address?

Kerndog73:
Even though the math for the 4/3 estimate is incredibly simple, I still think that this should be added to the library. Despite its simplicity, I had to look it up to make sure I didn't mess up. A named function is better than math. It would be nice if this was `const` too.

marshallpierce:
That's fair. Once I have a block of time to finish up the current in progress work (which may change its API a little) I'll make it public.

marshallpierce:
Done in https://github.com/marshallpierce/rust-base64/pull/167.

PR #152, opened 2021-01-21
Make DecoderReader own its inner reader

### LINKED ISSUE #151: Inconsistency between `EncoderWriter` and `DecoderReader` with 0.13
I had wrappers for decoders and encoders they can own their respective reader or writer and saw that 0.13 already changed this for encoders, but not for decoders. I was wondering if this is intentional, or could be changed for >=0.14?

rvolgers:
This makes it consistent again with `DecoderWriter`. Closes #151.

Name of the inner reader is changed to be slightly more descriptive.

Also adds an `into_inner` function because I need it for something I'm
working on. This is similar to `std::io::BufReader::into_inner`, which
also has a similar disclaimer.

### DIFF
--- a/src/encode.rs
+++ b/src/encode.rs
@@ -1,6 +1,6 @@
-use crate::{Config, PAD_BYTE};
 #[cfg(any(feature = "alloc", feature = "std", test))]
 use crate::{chunked_encoder, STANDARD};
+use crate::{Config, PAD_BYTE};
 #[cfg(any(feature = "alloc", feature = "std", test))]
 use alloc::{string::String, vec};
 use core::convert::TryInto;
--- a/src/read/decoder.rs
+++ b/src/read/decoder.rs
@@ -29,10 +29,10 @@ const DECODED_CHUNK_SIZE: usize = 3;
 /// assert_eq!(b"asdf", &result[..]);
 ///
 /// ```
-pub struct DecoderReader<'a, R: 'a + io::Read> {
+pub struct DecoderReader<R: io::Read> {
     config: Config,
     /// Where b64 data is read from
-    r: &'a mut R,
+    inner: R,
 
     // Holds b64 data read from the delegate reader.
     b64_buffer: [u8; BUF_SIZE],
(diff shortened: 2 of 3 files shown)


marshallpierce:
Thanks, I'll get this in the next release.

rvolgers:
Thanks!

There is one more thing actually.

For true consistency it would probably be better to also make a `EncoderWriter::into_inner`. However, there is already `EncoderWriter::finish` which gives the writer back (but only if no errors occurred), so I did not add it.

However, looking at this method in more detail, it is currently not a very pleasant situation. Its implementation is complicated because it does not consume `self` (comment says it''s because finish should be retry-able, which seems reasonable), and this even causes a need for extra checks in `write`.

I think it would be best to add `EncoderWriter::into_inner` and change the return value of `finish` into `Result<()>`, and clean up the now unnecessary checks for missing writer.

I'm happy to do this (if you agree of course) but don't know exactly if I can do it today.

marshallpierce:
I think finish has to dispose of the writer one way or another, though: we can't allow double finish(), or additional writes after finish(). If it doesn't return and consume the inner `W`, then it would need to stash it somewhere else so that into_inner() could return it, yet further writes would fail. That's even _more_ complex, but maybe it's worth it to offer an into_inner()? It could be documented that calling into_inner() after a successful finish() leaves the delegate writer in a known good state. Anyway we can hash out how the writer can be improved separately -- let's not delay this change for that.

rvolgers:
Hi, sorry to leave this hanging, I got a little distracted with other PRs.

The situation was indeed a little more tricky than it looked, and I pretty much agree with what you said.

I just submitted a minimal change which just adds an `into_inner` which can be used to recover the writer if `finish()` failed (or if the user just decides not to write any more data, but still wants the writer back for some reason). If there's no writer anymore it just panics, similar to what `finish()` does.

I think it should be okay like this, let me know if you disagree or if I missed something in your comment.

marshallpierce:
LGTM, thanks!

PR #153, opened 2021-02-03
Propose side-channel resistant feature

dingelish:
Hi Marshall,

Thanks for your great work and I really enjoy using your base64 crate :-) It's really fast!

We know that rustls has been widely adopted and it depends on `base64` to parse PEM files. Recently Apache httpd is constructing a new `mod_tls` based on rustls to replace the previous `mod_ssl` based on openssl. Glad to see rustls as well as base64 will be shipped with Apache httpd as well.

Recently I received a report from a group of security researchers (florian.sieck@uni-luebeck.de, s.berndt@uni-luebeck.de, thomas.eisenbarth@uni-luebeck.de), saying that this base64 decoder is not constant time thus suffer from side-channel attacks. The consequence is that PEM decoder and rustls suffers from side-channel attacks. Google has a good solution in BoringSSL: https://github.com/google/boringssl/blob/master/crypto/base64/base64.c . This proposed decoding algorithm is the researcher's work. They proved it in a unpublished paper. If you're interested, I can loop you in the mailing thread.

Now the solution is feature gated by `slow_but_safe` feature. `cargo test` passes. `cargo bench` shows performance slowdown. But it's acceptable if we only use this feature for PEM decoding (usually not a frequent operation). If this PR got merged, I'll create another PR for rustls.

Please feel free to edit/comment. Thanks!

Best, Yu

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ structopt = "0.3"
 default = ["std"]
 alloc = []
 std = []
+slow_but_safe = []
 
 [profile.bench]
 # Useful for better disassembly when using `perf record` and `perf annotate`
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -444,6 +444,18 @@ fn write_u64(output: &mut [u8], value: u64) {
     output[..8].copy_from_slice(&value.to_be_bytes());
 }
 
+#[cfg(feature = "slow_but_safe")]
+fn decode_aligned(b64ch: u8, decode_table: &[u8; 256]) -> u8 {
+    let mut result: u8 = 0x00;
+    let mut mask: u8;
+    let idx: [u8;2] = [ b64ch % 64, b64ch % 64 + 64];
+    for i in 0..2  {
+        mask = 0xFF ^ (((idx[i] == b64ch) as i8 - 1) as u8);
+        result = result | (decode_table[idx[i] as usize] & mask);
+    }
+    result
+}
+
 /// Decode 8 bytes of input into 6 bytes of output. 8 bytes of output will be written, but only the
 /// first 6 of those contain meaningful data.
 ///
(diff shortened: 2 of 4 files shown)


marshallpierce:
Thanks for this; a constant time implementation has been on my mind for quite some time! However, rather than a feature  approach, I'd prefer to go with making the "codec" part of base64 a generic type parameter so that the user could choose the current scalar implementation, your constant time scalar impl, and eventually AVX2 and NEON implementations. Most of the work would be in arranging the test corpus for the low-level encode and decode functions to be easily run for all implementations, and sprinkling the type parameter in all the right places. I think it would make sense to make the default for the codec be the current scalar impl, though I don't yet have a good sense of whether it would be a better API to pass around a struct for the codec (codec methods use &self) or just a type (no &self).

Would you have time to do that work?

dingelish:
@marshallpierce 
thanks for the quick reply! i'm willing to do it! and i think i'd be better to contribute after you have the generic type added. i'll definitely get this PR rebased then :-) thanks!

marshallpierce:
I'll have a draft PR up soon on trait-ifying the encoding and decoding, but in the mean time, can you confirm if the constant time decoding algorithm can have the same logic for the various `DecodeError` variants as the current way? If not, perhaps we should have different decoder implementations use different error types.

marshallpierce:
As an example of the above, will it do the same "bogus trailing bits" detection when decoding? If not, that would mean that it would need a different config type as well.

dingelish:
@marshallpierce thank you! i'd lookinto the "bogus trailing bits detection" part this weekend!

dingelish:
i just quickly look into the current trailing detection part. i think the proposed solution in this PR shall share the same bogus trailing bits detection with the current decoder :-)

marshallpierce:
Is the check for trailing bits insufficiently constant time, though, I wonder?

I'm making good progress on making pluggable "engines" so I should have a draft PR up for you to look at shortly to make sure it will fit with this proposed implementation.

dingelish:
@marshallpierce i kinda believe your current 'trailing bit detection' is almost identical to boringSSL's implemention:

https://boringssl.googlesource.com/boringssl/+/refs/heads/master/crypto/base64/base64.c#393

how do you think?

btw: i sent my PR to the researchers from University of Lübeck and they reviewed it and approved :-)

stormshield-gt:
@dingelish do you know where we can find the paper you mentioned about?

Issue #148, opened 2020-11-06
EncoderWriter::write_all leads to WriteZero error


paolobarbolini:
Using `base64::write::EncoderWriter` with a writer that accepts very limited writes leads to `EncoderWriter::write_all` calls failing.

## To reproduce ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ac6082fc272d0ecc767f7c2d026d184c))

```rust
extern crate base64; // 0.13.0

use std::io::{self, Write};

/// A Writer that limits how much can be written in a single call
struct ShortWriter<'a, W> {
    writer: &'a mut W,
    max_write_len: usize,
}

impl<'a, W> Write for ShortWriter<'a, W>
where
    W: Write,
{
    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
        let write_len = std::cmp::min(self.max_write_len, buf.len());
        let written = self.writer.write(&buf[..write_len])?;
        // do some other stuff...
        Ok(written)
    }

    fn flush(&mut self) -> io::Result<()> {
        self.writer.flush()
    }
}

fn main() {
    let contents = vec![b'A'; 100];

    let mut vec = Vec::with_capacity(8196);
    {
        let writer_ = ShortWriter {
            writer: &mut vec,
            max_write_len: 76,
        };
        let mut writer = base64::write::EncoderWriter::new(writer_, base64::STANDARD);
        writer.write_all(&contents).unwrap();
    }
    
    assert_eq!(vec, base64::encode(contents).into_bytes());
}
```

## To reproduce 2 (expanded the default Write::write_all implementation) ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3df150d7acf0b8ca538366bba231ec13))

<details>
  <summary>Click to expand!</summary>

```rust
extern crate base64; // 0.13.0

use std::io::{self, Write};

/// A Writer that limits how much can be written in a single call
struct ShortWriter<'a, W> {
    writer: &'a mut W,
    max_write_len: usize,
}

impl<'a, W> Write for ShortWriter<'a, W>
where
    W: Write,
{
    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
        let write_len = std::cmp::min(self.max_write_len, buf.len());
        let written = self.writer.write(&buf[..write_len])?;
        // do some other stuff...
        Ok(written)
    }

    fn flush(&mut self) -> io::Result<()> {
        self.writer.flush()
[...]

PR #161, opened 2021-06-23
Mention the write_all issue in the main EncoderWriter documentation

ajtribick:
As mentioned on #148.

Adds a note to check out the `Write` implementation documentation, which is hidden by default on docs.rs but contains useful information about the possibility for errors in `write_all`.


marshallpierce:
Thanks!

PR #162, opened 2021-06-27
Remove outdated comment

atouchet:
This comment is no longer relevant as the MSRV was bumped to 1.36.0 and criterion is no longer pinned to 0.3.2.


marshallpierce:
Thanks!

PR #187, opened 2022-04-30
decode vec size -- don't oversize the vector

marshallpierce:
Pointed out in #179.

Given that it's about to be resize()'d anyway, probably doesn't serve any purpose to allocate eagerly.


marshallpierce:
Offers a nice speed boost on tiny decodes.

PR #197, opened 2022-11-27
Update MSRV and criterion

marshallpierce:


marshallpierce:
nightly build is failing due to https://github.com/rust-lang/rust/issues/104815

Issue #206, opened 2022-12-14
Upgrade guide of `0.20` for users used to use config `base64::URL_SAFE_NO_PAD`


Xuanwo:
Users who used to use `base64::decode_config(&secret_key, base64::URL_SAFE_NO_PAD)` will be broken after upgrading from `0.13` to `0.20`. This issue intends to provide a quick upgrade guide.

Implement a engine:

```rust
/// The url safe engine for base64.
pub const BASE64_URL_SAFE_ENGINE: base64::engine::fast_portable::FastPortable =
    base64::engine::fast_portable::FastPortable::from(
        &base64::alphabet::URL_SAFE,
        base64::engine::fast_portable::NO_PAD,
    );
```

And replace all usage:

```diff
- base64::encode_config(&code_verifier_bytes, base64::URL_SAFE_NO_PAD)
+ base64::encode_engine(&code_verifier_bytes, &BASE64_URL_SAFE_ENGINE)
- base64::encode_config(&hasher.finalize(), base64::URL_SAFE_NO_PAD)
+ base64::encode_engine(&hasher.finalize(), &BASE64_URL_SAFE_ENGINE)
```

More info could be found at https://github.com/mozilla/sccache/pull/1477.

Xuanwo:
`RELEASE-NOTES` doesn't have enough information for this upgrade. How about adding this issue's content into `RELEASE-NOTES`? I'm willing to submit a PR for this.

Cc @marshallpierce

marshallpierce:
Thanks for the report. Based on feedback in https://github.com/marshallpierce/rust-base64/issues/205 I'll be quickly following 0.20 with 0.21, and part of that will be better documentation. I'll address this as part of that.

djmitche:
It would be great to have a complete migration guide, particularly since some of the behavior changes are a bit subtle.  Base64 strings containing url-unsafe characters are pretty unusual, and probably easy to miss in the unit tests of crates using this one.  And, a shift in expectations for padding on decoding might similarly be missed, if the test values are always correctly padded.

My understanding, confirmed with some experimentation:

| 0.13.1 Config   | 0.20.0 alphabet | `encode_padding` | `decode_padding_mode` |
|-----------------|-----------------|------------------|-----------------------|
| STANDARD        | STANDARD        | true             | Indifferent           |
| STANDARD_NO_PAD | STANDARD        | false            | Indifferent           |
| URL_SAFE        | URL_SAFE        | true             | Indifferent           |
[...]

Issue #205, opened 2022-12-12
No good alternative to `decode_config()` in 0.20


djc:
I'm unhappy about the API changes in 0.20. In my code base at work we have 12 use cases of base64 0.13 encoding and decoding, 9 of which pass a config (mostly `URL_SAFE_NO_PAD`, some `URL_SAFE`). Unless I misunderstand the new API (I don't see any examples of this use case in the README or the top-level documentation), in order to obtain the same result with 0.20 I would have to instantiate a custom `FastPortable` (which, despite being the only engine shipped with 0.20, I have to import from `base64::engine::fast_portable::FastPortable`) and initialize with `base64::alphabet::URL_SAFE` and `base64::engine::fast_portable::NO_PAD`.

Instead of offering a simpler way to do this through a top-level function, 6 of the 9 top-level functions now take an `Engine`; I would argue that, in case you already have an engine, these could have been methods attached to it.

I understand that the `Engine` abstraction is useful if you want to introduce a SIMD-backed implementation. I even like the introduction of a separate `Alphabet` type. I think it doesn't really make sense to have a `Config` type that doesn't include `Alphabet`, keeping it separate instead -- the whole point of having a `Config` type is to make it easy to wrap all the contextual/usually stable input for your main API calls into a single type. But to not offer a top-level API that makes it easy to use different alphabets/config seems to me like a lack of empathy for how this crate is used.

(FWIW, I also think jumping from 0.13 to 0.20 doesn't have much merit.)

marshallpierce:
First off, thanks for filing feedback. I released the 0.20 alpha precisely to gather feedback, but that didn't produce any... oh well. If 0.20's role is to be the lightning rod, that's ok! 
It is a bit baffling, though, to read what appears to be complaining about changing 12 callsites in a codebase. That is not a heavy burden. I don't really understand complaints about module paths, either. You can `use` them with whatever amount of prefix you want, and if your editor doesn't automatically do that for you, get one that does...? I do regret shipping with the `FastPortable` name, though -- I intended to change that before release but it slipped my mind.

I hadn't moved things to be methods on Engine to avoid perturbing the API too much, but I'm not opposed to it. 

I went back and forth during development about if `FastPortable`'s config should include an `Alphabet`, actually, and I think it's viable either way. However, if the goal is to have one thing that joins a `Config` with an `Alphabet`, isn't `Engine` that thing? You create it ahead of time and re-use it. It encapsulates all the behavior in one type.

[...]

Issue #192, opened 2022-08-01
base64::decode_config_slice should not panic


snyball:
I think the panicking behavior of `base64::decode_config_slice` is a mistake, this is an error that an application receiving untrusted input should be able to deal with. The size of decoded data can be calculated from the base64 input, before running `base64::decode_config_slice`;  but I think this goes against separation of concerns. The API user shouldn't have to know about internals like that.

There are situations where one wants to decode some untrusted base64 input, with a known maximum/absolute size, where the size is small enough that heap-allocation is unnecessary. A classic example of this would be a session ID.

## Suggested implementation

I think `DecodeError` should have a `BufferOverflow` variant, which `base64::decode_config_slice` should return when it overflows a given mutable slice. 

I foresee two use-cases for this error:
1. Intended: Abort handling whatever request submitted base64 with an invalid decode-size, and log/report the error without having to unwind.
2. Unintended: Increasing the size of the output buffer until the decoding succeeds
 
Use-case number 2 should be strongly discouraged by the API docs, with a reference to `base64::decode` for decoding base64 of unknown length.

## Backwards-compatibility
 
The suggested implementation is not backwards compatible because it changes `DecodeError`, and is therefore unrealistic given the widespread use of this crate in the ecosystem.

The implementation should therefore add a new function `basse64::decode_config_fixed_slice_bikeshed` with a new error-type `SliceDecodeErrorBikeshed` with variants `BufferOverflow`, and `DecodeError(DecodeError)` while deprecating `base64::decode_config_slice`.

marshallpierce:
- If we add another enum variant, it need only apply to `decode_config_slice`(now `decode_engine_slice`). That is perhaps an acceptable burden rather than introducing yet another function.
- Assuming we don't change the existing function: `decode_engine_slice_checked` to evoke `checked_add` and the like?
- In the meantime of course you can write your own wrapper function, but I realize that's not satisfying.
- To be useful, does the error have to be precise? That is, is it ok to require a size-3 decode buffer for input "AA==" that could decode to 3 bytes, but in fact only decodes to 1? This affects whether all write paths need to be aware of the slice size vs relying on panicking, or if a cheaper pre-calculation is acceptable.

snyball:
I don't think the error has to be precise beyond "doesn't fit," but then there could be cases where an API-user is not looking for "should be exactly this many bytes" but instead "can be maximum this amount of bytes." I haven't looked in to that option and have never been in that situation myself.

[...]

PR #207, opened 2022-12-15
API rework

marshallpierce:
- [x] Incorporate feedback from https://github.com/marshallpierce/rust-base64/issues/205
  - [x] Rename FastPortable
  - [x] Move encode and decode to methods
  - [x] Consts for common engine arrangements
  - [x] Better samples in top level docs
- [x] Doc improvements from https://github.com/marshallpierce/rust-base64/issues/206
  - [x] More details on padding changes (implicit indifferent)
- [x] Error instead of panicking from https://github.com/marshallpierce/rust-base64/issues/192

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,3 +1,49 @@
+# 0.21.0
+
+(not yet released)
+
+
+## Migration
+
+### Functions
+
+| < 0.20 function         | 0.21 equivalent                                                                     |
+|-------------------------|-------------------------------------------------------------------------------------|
+| `encode()`              | `engine::general_purpose::STANDARD.encode()` or `prelude::BASE64_STANDARD.encode()` |
+| `encode_config()`       | `engine.encode()`                                                                   |
+| `encode_config_buf()`   | `engine.encode_string()`                                                            |
+| `encode_config_slice()` | `engine.encode_slice()`                                                             |
+| `decode()`              | `engine::general_purpose::STANDARD.decode()` or `prelude::BASE64_STANDARD.decode()` |
+| `decode_config()`       | `engine.decode()`                                                                   |
+| `decode_config_buf()`   | `engine.decode_vec()`                                                               |
+| `decode_config_slice()` | `engine.decode_slice()`                                                             |
+
+The short-lived 0.20 functions were the 0.13 functions with `config` replaced with `engine`.
+
+### Padding
+
+If applicable, use the preset engines `engine::STANDARD`, `engine::STANDARD_NO_PAD`, `engine::URL_SAFE`,
+or `engine::URL_SAFE_NO_PAD`.
+The `NO_PAD` ones require that padding is absent when decoding, and the others require that
+canonical padding is present .
+
+If you need the < 0.20 behavior that did not care about padding, or want to recreate < 0.20.0's predefined `Config`s
+precisely, see the following table.
+
+| 0.13.1 Config   | 0.20.0+ alphabet | `encode_padding` | `decode_padding_mode` |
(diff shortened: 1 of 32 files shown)


RELEASE-NOTES.md
```diff
@@ -1,22 +1,90 @@
+# 0.20.1
```
djmitche, line 1: 0.21.0?

djmitche (commented):
Thanks for the updated release notes!

RELEASE-NOTES.md
marshallpierce: Oops, good catch

Gelbpunkt:
Hello @marshallpierce,

thanks for this update. I like the concise new API, but there seems to be somewhat of a regression (?) regarding the calculation of output in `decode_config_slice`.

I've put together an example [in this gist](https://gist.github.com/Gelbpunkt/39686a0f7cbd8f09a692ee8fe3531ae8). The quick rundown is that I have a websocket implementation that base64-encodes a slice of 20 bytes, which will then be 28 bytes large.

base64 0.13 was able to decode this into a slice with the length 20 just fine, even though it claims to panic if the slice is too small. base64 0.21-beta.2 will error because the length estimate for an input of length 28 returns 21 and it fails early.

I am guessing that this is due to the estimates being conservative, however there is no way for me to decode into a slice of length 20 now, even though it will work! Would it be possible to make the checking of the output length lazy (as I assume that must be how 0.13 did it, I haven't read the source), or adding a method that skips this check and panics instead?

marshallpierce:
> - The `*_slice` decode/encode functions now return an error instead of panicking when the output slice is too small
    - As part of this, there isn't now a public way to decode into a slice _exactly_ the size needed for inputs that
      aren't multiples of 4 tokens. If adding up to 2 bytes to always be a multiple of 3 bytes for the decode buffer is
      a problem, file an issue.
      
@Gelbpunkt I see this part has come true. :) I was persuaded in https://github.com/marshallpierce/rust-base64/issues/192 that `decode_slice` should not panic so that there was a safe way available. However, I don't mind adding back the old behavior, perhaps as `decode_slice_unchecked`. Would that work for you? It would be more work to make a more precise and/or lazy check, so I can explore that later, but for now I think just adding back the old logic is fine.

Gelbpunkt:
> > * The `*_slice` decode/encode functions now return an error instead of panicking when the output slice is too small
> >   - As part of this, there isn't now a public way to decode into a slice _exactly_ the size needed for inputs that
> >   aren't multiples of 4 tokens. If adding up to 2 bytes to always be a multiple of 3 bytes for the decode buffer is
> >   a problem, file an issue.
> 
> @Gelbpunkt I see this part has come true. :) I was persuaded in #192 that `decode_slice` should not panic so that there was a safe way available. However, I don't mind adding back the old behavior, perhaps as `decode_slice_unchecked`. Would that work for you? It would be more work to make a more precise and/or lazy check, so I can explore that later, but for now I think just adding back the old logic is fine.

I agree that not panicking is the better implementation, even for my usecase. It allows for handling malicious user input (or in this case HTTP requests that a header value that'll exceed the buffer). That's why I would prefer lazy checking in the long run, but for now adding back the old behaviour as an unchecked variant would be good enough already. Thank you for the quick response!

marshallpierce:
I've captured that in https://github.com/marshallpierce/rust-base64/issues/210. In the meantime, I'll release rc1 with decode_slice_unchecked.

marshallpierce:
If people don't use the release candidates, there's no point to publishing them -- they're there to gather feedback. :) 

Since there's been no feedback thus far on rc.1, I'll probably release it as 0.21.0 in the next few days.

PR #216, opened 2023-01-12
Make DecodeEstimate never panic

Nugine:
The calculations in `GeneralPurposeEstimate::new` can be transformed to another formula which never panics. The performance should be the same after optimization.

Do I need to adjust corresponding documentation?


marshallpierce:
Thanks! Even though surely nobody would ever hit the usize::MAX - 3 limit, it's nice to simply have the panic be impossible. I ended up coming up with a formula that produces slightly tighter assembly: https://github.com/marshallpierce/rust-base64/compare/mp/decode-estimate?expand=1

PR #217, opened 2023-01-17
Switch to non-overflowing decoded length formula

marshallpierce:
No change to benchmarks.

Motivated by the alternate approach in https://github.com/marshallpierce/rust-base64/pull/216/.

### DIFF
--- a/tests/encode.rs
+++ b/tests/encode.rs
@@ -26,10 +26,9 @@ fn encode_all_ascii() {
 fn encode_all_bytes() {
     let mut bytes = Vec::<u8>::with_capacity(256);
 
-    for i in 0..255 {
+    for i in 0..=255 {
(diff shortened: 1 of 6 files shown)


tests/encode.rs
```diff
@@ -26,10 +26,9 @@ fn encode_all_ascii() {
 fn encode_all_bytes() {
     let mut bytes = Vec::<u8>::with_capacity(256);
 
-    for i in 0..255 {
+    for i in 0..=255 {
         bytes.push(i);
     }
-    bytes.push(255); //bug with "overflowing" ranges?
```
mina86, line 32: let bytes: Vec<u8> = (0..=255).collect();

marshallpierce: Heh, good point!

PR #218, opened 2023-01-20
Simplify chunk_encoder

mina86:
Firstly, use `[T]::chunks` rather than doing index calculation
manually.

Secondly, get rid of max_input_length.  The function was overly
complicated for no reason.  After all, no matter whether engine uses
padding or not, N-byte output buffer can accommodate (N/4*3) input
bytes.

Encode benchmarks here are kind of all over.  Some improvements and
some regressions:

    encode/
               3   29.272 ns   -0.4%   No change
              50   54.162 ns  +12.4%   Regression
             100   95.488 ns   -7.8%   Improvement
             500   255.84 ns   +3.5%   Regression
            3072   1.3527 µs   +0.1%   No change
         3145728   1.4584 ms   +0.2%   Within noise
        10485760   4.8923 ms   +1.6%   Regression
        31457280   16.087 ms   -2.6%   Improvement

    encode_display/
               3   41.038 ns   -6.9%   Improvement
              50   70.877 ns   +7.2%   Regression
             100   91.190 ns   +1.1%   Within noise
             500   264.58 ns   -3.1%   Improvement
            3072   1.4515 µs   -3.8%   Improvement
         3145728   1.4239 ms   -0.2%   Within noise
        10485760   4.9768 ms   -0.7%   Within noise
        31457280   18.202 ms   -1.5%   Improvement

    encode_reuse_buf/
               3   20.264 ns  -23.1%   Improvement
              50   46.631 ns   +2.1%   Regression
             100   66.405 ns   -2.0%   Improvement
             500   243.52 ns   +3.5%   Regression
            3072   1.3871 µs   -1.2%   Within noise
         3145728   1.4278 ms   +0.8%   Within noise
        10485760   5.0530 ms   -1.6%   Improvement
        31457280   15.798 ms   +0.7%   No change

    encode_reuse_buf_stream/
               3   21.007 ns   +1.1%   Regression
              50   56.542 ns   -0.9%   No change
             100   75.214 ns   -1.4%   Improvement
             500   247.61 ns   -4.2%   Improvement
            3072   1.2836 µs   +0.1%   No change
         3145728   1.3177 ms   -0.5%   Within noise
        10485760   4.3843 ms   +0.3%   Within noise
        31457280   14.058 ms   -0.5%   Within noise

    encode_slice/
               3   11.859 ns   +0.1%   No change
              50   30.239 ns   +2.8%   Regression
             100   53.383 ns   -6.4%   Improvement
             500   207.90 ns   +0.2%   No change
[...]

PR #180, opened 2022-03-06
perf(encode): vec alloc & utf8 unchecked

AaronO:
Improves throughput by 10-20%

We avoid zeroing the allocated vec since we'll write into it and we don't need to check utf8 when building the String since all alphabets emit ascii/utf8

## Benches

```
❯ rustup run nightly cargo bench -- '^encode_small_input/encode/'
   Compiling base64 v0.20.0-alpha.1 (/Users/aaron/git/rust-base64)
    Finished bench [optimized + debuginfo] target(s) in 3.43s
     Running unittests (target/release/deps/base64-a91f049e61d3804b)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 105 filtered out; finished in 0.00s

     Running unittests (target/release/deps/benchmarks-0b92500c69ee3caa)
WARNING: HTML report generation will become a non-default optional feature in Criterion.rs 0.4.0.
This feature is being moved to cargo-criterion (https://github.com/bheisler/cargo-criterion) and will be optional in a future version of Criterion.rs. To silence this warning, either switch to cargo-criterion or enable the 'html_reports' feature in your Cargo.toml.

Gnuplot not found, using plotters backend
encode_small_input/encode/3
                        time:   [31.275 ns 31.304 ns 31.337 ns]
                        thrpt:  [91.300 MiB/s 91.394 MiB/s 91.480 MiB/s]
                 change:
                        time:   [-12.333% -12.192% -12.057%] (p = 0.00 < 0.05)
                        thrpt:  [+13.710% +13.884% +14.068%]
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
encode_small_input/encode/50
                        time:   [46.873 ns 46.923 ns 46.977 ns]
                        thrpt:  [1015.0 MiB/s 1016.2 MiB/s 1017.3 MiB/s]
                 change:
                        time:   [-10.129% -9.9667% -9.8193%] (p = 0.00 < 0.05)
                        thrpt:  [+10.889% +11.070% +11.271%]
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  2 (2.00%) low mild
  3 (3.00%) high mild
  2 (2.00%) high severe
encode_small_input/encode/100
                        time:   [62.462 ns 62.484 ns 62.520 ns]
                        thrpt:  [1.4896 GiB/s 1.4905 GiB/s 1.4910 GiB/s]
                 change:
[...]

PR #224, opened 2023-02-05
Simplify chunk_encoder, take 2

marshallpierce:
Based on #218.


mina86:
FYI, my next step would be to get rid of add_padding actually.  This is because my long-term plan is to add support for MaybeUninit which would allow to do #179 and #180 soundly (and getting rid of add_padding makes my current code somewhat simpler).  See https://github.com/mina86/rust-base64/tree/master

marshallpierce:
I see; sounds interesting. Until that's ready, though, I'd prefer to keep add_padding so that it's all consistent until it can all be removed.

Issue #226, opened 2023-02-10
DecoderReader accepts incorrect input


mina86:
When parts read from underlying reader include valid padding the
reader accepts the input even if there’s more data to follow.  In this
example the reader produces string `AA==AA==AA==` which should result
in an error but instead generates three zero bytes.

```rust
struct Foo(usize);

impl std::io::Read for Foo {
    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
        if self.0 == 0 {
            return Ok(0);
        }
        self.0 -= 1;
        buf[..4].copy_from_slice(b"AA==");
        Ok(4)
    }
}

fn main() {
    use std::io::Read;
    use base64::engine::general_purpose::STANDARD;

    let mut decoder = base64::read::DecoderReader::new(Foo(3), &STANDARD);
    let mut vec = Vec::new();
    let res = decoder.read_to_end(&mut vec);
    print!("{res:?} {vec:?}");
}
```

marshallpierce:
Good find, thanks for the heads up.

marshallpierce:
This will be fixed in 0.21.1 momentarily.

Issue #190, opened 2022-07-21
FR: Provide decode_inplace function


mina86:
Would be nice to have an in-place decode function, i.e.:

```rust
fn decode_inplace(data: &mut [u8]) -> Result<usize, DecodeError> { … }
```

Since output is never longer than input, the decoded value can be saved directly into the input buffer thus possibly saving on having to allocate a new output buffer, e.g.:

```rust
fn b64decode(value: String) -> Result<Vec<u8>, base64::DecodeError> {
    let mut vec = value.into_bytes();
    let new_length = base64::decode_inplace(&mut vec[..])?;
    vec.truncate(new_length);
    Ok(vec)
}
```

This would of course come with a caveat that if error is encountered the data in the buffer is in unspecified state (portion of it could have been overwritten).

Nugine:
<https://docs.rs/base64-simd/latest/base64_simd/struct.Base64.html#method.decode_inplace>

```rust
fn b64decode(value: String) -> Result<Vec<u8>, base64_simd::Error> {
    let base64 = base64_simd::Base64::STANDARD;
    let mut vec = value.into_bytes();
    let new_length = base64.decode_inplace(&mut vec)?.len();
    vec.truncate(new_length);
    Ok(vec)
}
```

marshallpierce:
I can see the appeal in general of not requiring more memory footprint than necessary. Do you have a specific use case in mind? Does this functionality need to extend to anything beyond a function that decodes in place in order to be useful?

mina86:
There are two cases:
1. I have a String which I want to decode into Vec<u8>.  With decode_inline I can do it with zero allocations reusing String’s internal Vector.
2. I have an encoded data which (after decoding) is further deserialised into actual object I want.  Though to be honest I cannot find an example of that now and I’m not 100% sure this is still an issue for me.

Function decoding in place would be sufficient.

By the way, I realised that there is one complication.  If I understand base58 correctly, decoding happens from the end so the first byte that is known is the last byte in the buffer.  This means that decode_inplace would store the data at the end of the buffer.  This would complicate the first use case and require the use of Vec::copy_within.  Still, that’s probably better than having to allocate a new Vector.

marshallpierce:
Not sure about base58, but base64 decodes in front to back order in groups of 4 tokens, so that problem at least does not apply.

mina86:
[...]

PR #227, opened 2023-02-10
decode: use exact decoded length rather than estimation

### LINKED ISSUE #210: Precise length checking in `decode_slice`
It would be good to have the length error only emitted when we know it to be required.

### LINKED ISSUE #212: 0.21.0: decode_slice: Unexpected behavior for exact-size output slice
I noticed some behavior with `STANDARD_NO_PAD` that I didn't expect (similarly for `URL_SAFE_NO_PAD`). In the code below, I encode a 32 byte string, then use `decode_slice()` to read it into a buffer. 

```rust
use rand::prelude::*;
use base64::{
    DecodeSliceError,
    engine::{Engine, general_purpose::STANDARD_NO_PAD as ENC}
}; // 0.21.0

fn main() {
    let x = thread_rng().gen::<[u8; 32]>();
    
    // Decode into a 32 byte buffer. I would expect this to succeed, since the
    // encoding is not padded.
    let mut buf = [0_u8; 32];
    assert!(matches!(
        ENC.decode_slice(ENC.encode(x), &mut buf[..]),
        Err(DecodeSliceError::OutputSliceTooSmall),
    ));
    
    // Decode into a 64 byte buffer. I would expect this to be longer than
    // necessary.
    let mut buf = [0_u8; 64];
    ENC.decode_slice(ENC.encode(x), &mut buf[..]).unwrap();
    assert_eq!(x, buf[..32]);
    
    // For what it's worth, the unchecked version does not panic if the buffer
    // holds only 32 bytes.
    let mut buf = [0_u8; 32];
    ENC.decode_slice_unchecked(ENC.encode(x), &mut buf[..]).unwrap(); // ok
    assert_eq!(x, buf);
}
```

Is this behavior expected? If so, is it documented anywhere?

mina86:
Fixes: https://github.com/marshallpierce/rust-base64/issues/210
Fixes: https://github.com/marshallpierce/rust-base64/issues/212

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -1,4 +1,4 @@
-use crate::engine::{general_purpose::STANDARD, DecodeEstimate, Engine};
+use crate::engine::Engine;
 #[cfg(any(feature = "alloc", feature = "std", test))]
 use alloc::vec::Vec;
 use core::fmt;
@@ -89,7 +89,7 @@ impl From<DecodeError> for DecodeSliceError {
 #[deprecated(since = "0.21.0", note = "Use Engine::decode")]
 #[cfg(any(feature = "alloc", feature = "std", test))]
 pub fn decode<T: AsRef<[u8]>>(input: T) -> Result<Vec<u8>, DecodeError> {
-    STANDARD.decode(input)
+    crate::engine::general_purpose::STANDARD.decode(input)
 }
 
 /// Decode from string reference as octets using the specified [Engine].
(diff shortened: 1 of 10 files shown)


mina86:
By the way, I’m beginning to think that the interface for the internal_decode should be *completely* changed.  Specifically to something like:

```rust
fn internal_decode(
    dst: *mut [MaybeUninit<u8>],
    src: *const [u8],
) -> Result<(), DecodeError>;
```

With the following contract:
* `(*dst).len() / 3 == (*src).len() / 4`,
* `(*dst).len() % 3 == (*src).len() % 4 == 0`,
* source has no padding bytes and
* `dst` and `src` either do not overlap, or `dst as *mut () == src
  as *const ()`.

This would help address the following issues:
* https://github.com/marshallpierce/rust-base64/issues/190,
* https://github.com/marshallpierce/rust-base64/issues/210,
* https://github.com/marshallpierce/rust-base64/issues/212,
* https://github.com/marshallpierce/rust-base64/issues/226 and make those optimisations sound:
* https://github.com/marshallpierce/rust-base64/pull/179 and
* https://github.com/marshallpierce/rust-base64/pull/180.

tgross35:
@mina86 any chance you are able to push this over the line? The author suggested in https://github.com/marshallpierce/rust-base64/issues/210#issuecomment-1799038820 that maybe this could be split up into smaller PRs.

> By the way, I’m beginning to think that the interface for the internal_decode should be _completely_ changed. Specifically to something like:
> 
> ```rust
> fn internal_decode(
>     dst: *mut [MaybeUninit<u8>],
>     src: *const [u8],
> ) -> Result<(), DecodeError>;
> ```

FYI this would need to be an `unsafe` function, and `MaybeUninit` doesn't add anything to integers (since all possible byte representations are valid integers)

marshallpierce:
I'm quite happy to rearrange APIs to have nice new features, but `unsafe` is going to be a pretty hard sell. It adds quite a burden on users who are in environments that require auditing all `unsafe`.

marshallpierce:
Closing as stale

PR #231, opened 2023-03-31
Replace internal buffer in decoder with `BufRead`

### LINKED ISSUE #230: `DecoderReader` probably should accept `BufRead` instead of `Read`
I noticed `DecoderReader` holds its internal buffer of bytes. This has a number of disadvantages:

* Needless copying from in-memory readers (slices)
* Double-buffering already-buffered readers
* The bytes are lost when dropping or calling `into_inner`
* Reinvents the wheel; there already is `std::io::BufReader`

Unless I'm missing something this can be easily fixed just by requiring `BufRead` instead of `Read` and using it. It's a breaking change so it's better to do it before 1.0.

Related: maybe it itself could provide `BufRead` but I'm not sure if 3-byte buffer is fine or not.

Kixunil:
The `DecoderReader` used an internal buffer for reading which came with a number of disadvantages, such as:

* Needless copying from in-memory readers (slices)
* Double-buffering already-buffered readers
* The bytes are lost when dropping or calling `into_inner`
* `BUF_SIZE` is not configurable
* `std::io::BufReader` has access to some unstable optimizations which this crate cannot use
* Reinvents the wheel; there already is `std::io::BufReader`

This change removes it and requires `BufRead` instead. Decoding is implemented on top of `fill_buf` for larger chunks with fallback to small, stack-allocated buffer for tiny chunks that may appear at boundaries.

To resolve borrowing problems `decode_to_buf` had to be removed which also enabled decoding bytes directly into internal buffer when the buffer to be filled is small rather than into a temporary buffer which is then copied.

This improves performance of reading by around 7-22% on my machine.

Closes #230

### DIFF
--- a/examples/base64.rs
+++ b/examples/base64.rs
@@ -1,5 +1,5 @@
 use std::fs::File;
-use std::io::{self, Read};
+use std::io::{self, BufRead, BufReader};
 use std::path::PathBuf;
 use std::process;
 use std::str::FromStr;
@@ -48,7 +48,7 @@ struct Opt {
 fn main() {
     let opt = Opt::from_args();
     let stdin;
-    let mut input: Box<dyn Read> = match opt.file {
+    let mut input: Box<dyn BufRead> = match opt.file {
         None => {
             stdin = io::stdin();
             Box::new(stdin.lock())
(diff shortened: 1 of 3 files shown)


marshallpierce:
Thanks for the contribution. It will probably take me some time to review it, but I'll get to it when I can.

apasel422:
Any chance this could be rebased and reconsidered for merging?

Kixunil:
I'll be happy to rebase if @marshallpierce has the time to do review of this. Otherwise rebasing would be just a waste of time for me.

marshallpierce:
Please rebase and I will make time for it.

Issue #232, opened 2023-04-13
make Alphabet::from_str_unchecked public


lovasoa:
# The problem

- Alphabet::new(...).unwrap() cannot be used in const contexts
- some base64 implementations in the wild do not respect the rules implemented in Alphabet::new
  - <small>In particular, I am interested in the alphabet used in [Google Arts & Culture](https://artsandculture.google.com/) image time tokens in order to implement them in [dezoomify-rs](https://github.com/lovasoa/dezoomify-rs/blob/72ec0d92df09593ab5d4c50c5762d7c3e392ba9b/src/google_arts_and_culture/url.rs#L26)</small>

Making `Alphabet::from_str_unchecked` public would help me with that

marshallpierce:
What does the image time token alphabet need? 

Also, would lazy_static serve while we wait for unwrap to be const? In fact, static might be _preferable_ so that only one copy of the alphabet needs to exist in memory.

Dushistov:
> Alphabet::new(...).unwrap() cannot be used in const contexts

Actually it can, sort of. `panic!` can be used in const context,
so you can write:

```
use base64::alphabet::Alphabet;

static CUSTOM: Alphabet = {
    match Alphabet::new("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/") {
        Ok(x) => x,
        Err(_) => panic!("creation of alphabet failed"),
    }
};
```

and it compiles just fine with current stable rustc.

marshallpierce:
lazy_static works fine:

```
use base64::{
    alphabet::Alphabet,
    engine::{general_purpose::GeneralPurpose, GeneralPurposeConfig},
};

lazy_static::lazy_static! {
    static ref CUSTOM: Alphabet = Alphabet::new("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/").unwrap();
}

fn main() {
    let engine = GeneralPurpose::new(&CUSTOM, GeneralPurposeConfig::new());
}
```

Between that and the const/panic workaround, it doesn't seem worth the risk of opening up from_str_unchecked.

PR #234, opened 2023-04-20
make Alphabet::from_str_unchecked public

Phil-hacker:
fixes issue #232


marshallpierce:
I don't think this is needed anymore since there's a solution for the original issue. Thanks though.

PR #238, opened 2023-05-21
Fix DecoderReader handling of padding

### LINKED ISSUE #226: DecoderReader accepts incorrect input

marshallpierce:
Fixes #226.

DecoderReader now keeps more state so that erroneous internal padding that happens to fall right on a read buffer boundary isn't silently ignored.


danielabrozzoni:
Hey, is there any reason why this PR should update the crate MSRV? We're using `base64` in https://github.com/bitcoindevkit/rust-esplora-client and we'd like to keep 1.57.0 as our MSRV, but at the same time we would prefer not to pin base64 to "0.21.0" as we'd like to have this bug fix in. Thanks!

marshallpierce:
Huh, I figured that was old enough most people won't care, but I can roll that back. 1.60 is needed to build the dev dependencies, but for the main code, 1.57 is still fine. I'll release another version with that change.

danielabrozzoni:
That would be great! Thanks :)

marshallpierce:
Fix incoming in https://github.com/marshallpierce/rust-base64/pull/239

PR #241, opened 2023-06-20
Optimize generic functions

EFanZh:
This PR intends to reduce binary size by sharing codes between different `AsRef<[u8]>` types.

### DIFF
--- a/src/engine/mod.rs
+++ b/src/engine/mod.rs
@@ -114,14 +114,23 @@ pub trait Engine: Send + Sync {
     ///
     /// let b64_url = CUSTOM_ENGINE.encode(b"hello internet~");
     #[cfg(any(feature = "alloc", feature = "std", test))]
+    #[inline]
     fn encode<T: AsRef<[u8]>>(&self, input: T) -> String {
-        let encoded_size = encoded_len(input.as_ref().len(), self.config().encode_padding())
-            .expect("integer overflow when calculating buffer size");
-        let mut buf = vec![0; encoded_size];
+        fn inner<E>(engine: &E, input_bytes: &[u8]) -> String
+        where
+            E: Engine + ?Sized,
+        {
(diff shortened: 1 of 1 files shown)


marshallpierce:
How are you measuring binary size? How much of a difference does this change make?

EFanZh:
I just compared the sizes of the result binaries.

For example, this code:

```rust
use base64::engine::general_purpose;
use base64::Engine;
use std::hint;

fn use_base64(input: &(impl AsRef<[u8]> + ?Sized)) {
    let input = hint::black_box(input);

    hint::black_box(general_purpose::STANDARD.encode(input));
    hint::black_box(general_purpose::STANDARD.decode(input)).ok();
}

fn main() {
    use_base64(&[]);
    use_base64(&[0]);
    use_base64(&[0, 1]);
    use_base64(&[0, 1, 2]);
    use_base64(&[0, 1, 2, 3]);
    use_base64(&[0, 1, 2, 3, 4]);
    use_base64(&[0, 1, 2, 3, 4, 5]);
    use_base64(&[0, 1, 2, 3, 4, 5, 6]);
    use_base64(&[0, 1, 2, 3, 4, 5, 6, 7]);
}
```

On my PC, my PR will reduce the size of the result executable from 172032 bytes to 159232 bytes on release profile, which is 12.5 KiB.

marshallpierce:
I see. This isn't exactly a terrible burden code wise, but it is somewhat awkward. Is this a problem in practice?

EFanZh:
Yes, I have a project that suffers from binary size bloating, and after some investigation, I have found that part of the problem is cause by different `base64` generic function instances.

Also, the method I am using is a fairly standard pattern: <https://www.possiblerust.com/pattern/non-generic-inner-functions>. The standard library also uses this pattern: <https://doc.rust-lang.org/1.70.0/src/std/fs.rs.html#249-258>.

marshallpierce:
OK, makes sense. I'll merge and release this shortly.

PR #243, opened 2023-07-12
Implement `source` instead of `cause` on std::error::Error types

bartelsielski:
The `cause` method has been deprecated since Rust 1.33.0 and has been replaced by `source`.  The `cause` method's default implementation calls `source` so there is no need to implement them both.

On types which always return `None` in `source`, the implementation has been removed because the default implementation does exactly this.


marshallpierce:
Thanks for the contribution. I'll be releasing a new version shortly, and this will be included.

PR #246, opened 2023-08-10
Roll back the MSRV to Rust 1.48

tcharding:
The current MSRV (for non-dev builds) is Rust 1.57 however the codebase does not use any features from edition 2021 (landed Rust 1.56) except the `TryFrom` trait.

If we explicitly import `core::convert::tryFrom` we can stay on edition 2018.

The `rust-version` manifest option was added in Rust 1.56 but lets leave it there as added documentation of the MSRV and in preparation for bumping the MSRV when needed.

Rust 1.48 is a suitably old toolchain version, it is the toolchain included in Debian Bullseye.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ workflows:
                 # be easier on the CI hosts since presumably those fat lower layers will already be cached, and
                 # therefore faster than a minimal, customized alpine.
                 # MSRV
-                'rust:1.57.0'
+                'rust:1.48.0'
               ]
               # a hacky scheme to work around CircleCI's inability to deal with mutable docker tags, forcing us to
               # get a nightly or stable toolchain via rustup instead of a mutable docker tag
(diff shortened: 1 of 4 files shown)


tcharding:
I was going to create an issue but I saw the very amusing "How I, the issue filer, am going to help solve it" :)

tcharding:
PR currently uses `use std::convert::TryFrom;`, would you prefer `use std::convert::TryFrom as _;` as it was in the tests? I don't know the exact benefit, is it to future proof against name clashes?

marshallpierce:
If planned obsolescence was taken as a motivating goal, the result would be Debian...

I'll take care of the trait tweaking. The _ thing is to avoid being able to accidentally refer to things by their "bare" name, in keeping with the habit I picked up from working on [ring](https://github.com/briansmith/ring) of leaving a bit of the module path intact when working directly on types from other crates.

tcharding:
Thanks man!

PR #253, opened 2023-09-16
Simplify `cfg`s and allow testing without default features.

kpreid:
This PR contains two commits which address different matters, but the second depends on and helps test the first. If desired, I can split it into two PRs, or discard either one and keep the other.

---

First: `no_std` import simplification.

It's always OK to use `core` and `alloc` if you're using `std`, and `core` if you're using `alloc`. Therefore, code can be simplified by using `alloc` even if `std` is also being used, and so on.

* Make `std` feature depend on `alloc` feature.
* Simplify `cfg()`s by making that assumption.
* Don't import `std` aliased as `alloc`; use `alloc` directly.

These changes do not affect the public API of the crate in any way, but should make maintenance slightly easier since there are fewer subtly-different conditions under which the code might be compiled.

---

Second: Allow `cargo test` to complete under any set of features. 

Previously, tests, examples, and benches would fail to compile under `cargo test --no-default-features`. Now they are all skipped if they would not compile.

This does not affect behavior *with* default features — test coverage is not reduced.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,19 @@ rust-version = "1.48.0"
 [[bench]]
 name = "benchmarks"
 harness = false
+required-features = ["std"]
+
+[[example]]
+name = "base64"
+required-features = ["std"]
+
+[[test]]
+name = "tests"
+required-features = ["alloc"]
+
+[[test]]
+name = "encode"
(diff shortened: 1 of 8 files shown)


marshallpierce:
Thanks for the cleanup!

PR #256, opened 2023-10-23
Fix missing verb in the crate documentation

nicholasbishop:


marshallpierce:
Thanks for the contribution. I ended up doing https://github.com/marshallpierce/rust-base64/pull/260 instead, as I don't really think of base64 as a "format" per se.

PR #258, opened 2023-11-08
Update example to use clap 4, replace lazy_static with once_cell, update other dependencies

serprex:
structopt crate was deprecated after being merged into clap v3


marshallpierce:
Thanks for the updates. I'll take it from here in https://github.com/marshallpierce/rust-base64/pull/259.

marshallpierce:
Unfortunately most of the updates weren't viable with a 1.48 MSRV. Thanks anyway.

PR #261, opened 2023-12-04
Update documentation for v0.21 deprecations

silverlyra:
👋🏻 Hello! I used this `base64` crate for the first time this past week in a project. Honestly when I started, I first didn’t even look at the documentation – I ran a `cargo add base64`, and then tried typing `base64::decode` in my editor.

That function did exist, and does exactly what I wanted (decode canonical base64 input using the standard alphabet), but it showed as ~deprecated~. So I went to docs.rs to see what I was supposed to use instead, and I found the current crate documentation difficult to understand how to set up base64 decoding and/or encoding.

I appreciate this crate and the flexibility it offers for the nuances of base64 (different alphabets, padding `=` character handling), and here I’ve made an attempt to update the crate documentation to help a new user navigate those nuances, and successfully start decoding/encoding data.

side by side base64 rustdoc

In the updated documentation, I’ve made sure there’s a code sample immediately visible which shows how to do decoding and encoding using default settings, and I (attempt to) explain enough of the nuance of base64 to empower users who would want non-default choices to get the setup they need.

I’ve submitted many more pull requests on code than on prose 😅 – I’m not sure how you’d want to review this, but I can keep workshopping the text. 🙏🏻 I also made some tweaks to the existing `base64.rs` example.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ rustdoc-args = ["--generate-link-to-definition"]
 criterion = "0.4.0"
 rand = { version = "0.8.5", features = ["small_rng"] }
 structopt = "0.3.26"
+strum = { version = "0.25", features = ["derive"] }
 # test fixtures for engine tests
 rstest = "0.13.0"
 rstest_reuse = "0.6.0"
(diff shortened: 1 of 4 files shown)


marshallpierce:
Thanks very much for the thoughtful improvements! I've taken care of the conflicts (#263) and tweaked a couple of minor things (notably, not renaming the engine imports to BASE64 to highlight to the hasty reader that they are in fact different).

PR #262, opened 2024-01-07
examples: replace structopt with clap

tamird:
See https://github.com/TeXitoi/structopt/issues/525.

Currently base64 is the most downloaded dependent of structopt:
https://crates.io/crates/structopt/reverse_dependencies.


marshallpierce:
Thanks! It's a shame the ancient-MSRV crowd forces the use of clap 3, but I'm glad to help structopt's retirement.

PR #264, opened 2024-01-11
provide as_string() method to return the alphabet characters

JosiahParry:
This PR provides a method to return the characters in the `Alphabet` struct as a `String`. 

At present, there is no provided way to retrieve characters that were used to create the alphabet. One way to accomplish this today, is to capture the output of the `Debug` format and process it manually. This `as_string()` method will make it easier to do just that. Alternatively, if the `symbols` field of the struct was made public, the onus could be put on the users of the crate to iterate through the array and cast u8 as char.

As an alternative to a GH issue, I wanted to provide a PR instead. Please let me know if this suffices.

### DIFF
--- a/src/alphabet.rs
+++ b/src/alphabet.rs
@@ -124,6 +124,7 @@ impl Alphabet {
         Ok(Self::from_str_unchecked(alphabet))
     }
 
+    /// Create a String from the symbols in the `Alphabet`
     pub fn as_string(&self) -> String {
         self.symbols.iter().map(|c| *c as char).collect::<String>()
     }
--- a/src/alphabet.rs
+++ b/src/alphabet.rs
@@ -123,6 +123,10 @@ impl Alphabet {
 
         Ok(Self::from_str_unchecked(alphabet))
     }
+
+    pub fn as_string(&self) -> String {
+        self.symbols.iter().map(|c| *c as char).collect::<String>()
+    }
 }
 
 impl convert::TryFrom<&str> for Alphabet {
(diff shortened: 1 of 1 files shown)


JosiahParry:
Firs time contributing to something with `no_std` I _think_ i got the cfg correct :P

src/alphabet.rs
```diff
@@ -123,6 +123,12 @@ impl Alphabet {
 
         Ok(Self::from_str_unchecked(alphabet))
     }
+
+    /// Create a String from the symbols in the `Alphabet`
+    #[cfg(any(feature = "std", test))]
+    pub fn as_string(&self) -> String {
+        self.symbols.iter().map(|c| *c as char).collect::<String>()
```
marshallpierce, line 130: This could avoid allocation, and thus be no_std compatible, by returning a &str: use std::str::from_utf8 on symbols. It's safe to unwrap/expect that, since symbols are verified to be printable ascii. Users in std environments can then .to_string() the str if needed.

marshallpierce (changes requested):
Thanks for the contribution! I've suggested a simplification (and it looks like you did get the attributes right, but you won't need them after all).

src/alphabet.rs
JosiahParry: Ooh interesting! Would this look something like `std::str::from_utf8(self.symbols.as_ref()).unwrap()`

JosiahParry: If that's the case, I suppose the method should be more aptly named `as_str()`

marshallpierce: Yep, that should do it.

JosiahParry: Perfect. Pushed a new commit. Resolving!

src/alphabet.rs
```diff
@@ -123,6 +123,11 @@ impl Alphabet {
 
         Ok(Self::from_str_unchecked(alphabet))
     }
+
+    /// Create a `&str` from the symbols in the `Alphabet`
+    pub fn as_str(&self) -> &str {
+        std::str::from_utf8(&self.symbols).unwrap()
```
marshallpierce, line 129: oh right... CI is mad because you can't use `std` in `no_std`. Just `str`, or the path using `core`, should work.

JosiahParry:
CI Gods are please 👼🏼

marshallpierce (approved):

marshallpierce:
Released as 0.21.7

PR #269, opened 2024-03-02
Precise decode output slice length checking

marshallpierce:


marshallpierce:
Fixes https://github.com/marshallpierce/rust-base64/issues/210

PR #273, opened 2024-07-10
Documentation example for constant-space validity checking

alexpovel:
Methods like `Engine::decode` require allocations
sized to fit all output. For use cases where
(potentially large) base64 input merely needs to
be validated as correctly shaped (alphabet,
padding), this approach relaxes memory
requirements.


marshallpierce:
Thanks!

PR #274, opened 2024-10-05
Add a lot of changes that Clippy suggested when enabling some lints

nandolawson:
None of the changes affect the behavior of the code


martinfrances107:
I have visually scanned all the changes ... everthing looks as expected

marshallpierce:
Thanks for the contribution!

Issue #275, opened 2024-11-18
Default issue


rkuklik:
Hello. I was looking at `base64::write::EncoderStringWriter`, which requires `StrConsumer`. However, it is only implemented for `String` while looking basically as `core::fmt::Write`.

Could I either provide blanket `impl<W: Write> StrConsumer for W` (which shouldn't be breaking) or rip it out and switch to `Write` trait directly (breaking)?

rkuklik:
On second thought, both options would be breaking, but `base64` isn't at vesion 1, so I would prefer trait replacement.

marshallpierce:
The two traits do bear a superficial similarity, but they do have a different intent to communicate for the user: `StrConsumer` is _just_ for base64, whereas `fmt::Write` is more general purpose (any Unicode), so I don't think it should just use `fmt::Write` directly. However, I agree that it should be possible for a user to use a `fmt::Write` as a `StrConsumer`. 

I doubt a breaking change in this specific case is likely to affect much (any) calling code, so I'm not very concerned about that aspect, but I still prefer an explicit user choice for the purposes of confirming the user's intent to avoid accidentally using a type in an unexpected way. This could be done by making an adapter struct `WriteConsumer<W>` (or whatever). Then, if you want to use a Write as a StrConsumer, it's just a `WriteConsumer::new(the_write)` away.

marshallpierce:
On the other hand, because there is only really one way that a Write could implement StrConsumer, maybe it _is_ better to just do the blanket impl. Hmm... Competing priorities is what makes API design interesting.

rkuklik:
> The two traits do bear a superficial similarity, but they do have a different intent to communicate for the user: `StrConsumer` is _just_ for base64, whereas `fmt::Write` is more general purpose (any Unicode), so I don't think it should just use `fmt::Write` directly. However, I agree that it should be possible for a user to use a `fmt::Write` as a `StrConsumer`.

I don't really see the difference in intent. Both represent types which can consume UTF-8 encoded bytes, while `StrConsume` doesn't offer any new capabilities or requirements (like `Eq` and `PartialEq` have).

[...]

PR #277, opened 2024-11-22
Rework UTF-8 writing

rkuklik:
See #275.

This needs more polish, mainly around docs and naming things. Also, the `into_inner` method is inconsistent with previous implementation, so it should either be removed or adjusted.


src/write/encoder_utf8.rs
```diff
@@ -0,0 +1,187 @@
...
+        Self::new(Utf8Compat::new(writer), engine)
+    }
+
+    /// Encode all remaining buffered data, including any trailing incomplete input triples and
+    /// associated padding.
+    ///
+    /// Returns the base64-encoded form of the accumulated written data.
+    pub fn formatter(mut self) -> W {
+        self.finish()
+            .expect("Writing to a consumer should never fail")
```
marshallpierce, line 101: This is what's not going to work for me. We should not assume that this will always succeed. Anyone is free to implement Write and take the contract at its word that write_str can fail.

marshallpierce:
Closing as stale

PR #279, opened 2025-01-01
Clarify the docs of `NO_PAD`, add shorthands for indifferent decoding.

de-vri-es:
This PR tweaks the documentation of the `general_purpose::NO_PAD` configuration. The existing wording could be interpreted as "does not require padding when decoding" instead of "requires that there is no padding when decoding".

The PR also adds indifferent variants of the padding configs. I believe that in most applications, that is the correct behaviour: be strict in what you generate, be lenient in what you accept.

I didn't add any shorthands for the engines, as that would mean adding 4 new constants in total (`STANDARD_PAD_INDIFFERENT`, `STANDARD_NO_PAD_INDIFFERENT`, `URL_SAFE_PAD_INDIFFERENT` and `URL_SAFE_NO_PAD_INDIFFERENT`).

Personally, I would like to add those too, since I think `[NO_]PAD_INDIFFERENT` is most often the correct choice. However, I wanted to check here if adding those would be appreciated first.

### DIFF
--- a/src/engine/general_purpose/mod.rs
+++ b/src/engine/general_purpose/mod.rs
@@ -347,11 +347,21 @@ pub const URL_SAFE_NO_PAD: GeneralPurpose = GeneralPurpose::new(&alphabet::URL_S
 
 /// Include padding bytes when encoding, and require that they be present when decoding.
 ///
-/// This is the standard per the base64 RFC, but consider using [`NO_PAD`] instead as padding serves
+/// This is the standard per the base64 RFC, but consider using [`NO_PAD_INDIFFERENT`] instead as padding serves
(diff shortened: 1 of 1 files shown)


marshallpierce:
Thanks. I'll make one small tweak to suggest _either_ N_P or N_P_I since for internal systems, users may control all participants and want to take advantage of the lack of `=`.

marshallpierce:
Oh, and as for the engines... perhaps adding the constants will help people write systems that at least tolerate the lack of padding. Anything that helps lessen the occurrence padding, which I hate with the passion of a thousand burning suns, is a good thing. So, I'll sprinkle those in too.

Issue #286, opened 2025-06-18
Check valid encoding given an alphabet


tcerqueira:
# The problem

- I want to check the encoding is valid without decoding

A bit follow-up to this #191 .

Given an alphabet this seems a useful feature to me. Currently, I'm doing as suggested: decoding, checking for errors and discarding the result but I'm not sure the compiler can always optimize the allocations away.

This is not complex, right? We can check for the 4 kinds of error in `DecodeError` without decoding.
- Character not in alphabet
- Invalid length
- Invalid symbol before padding
- Invalid padding

Asking for some feedback, I can work on it.

marshallpierce:
Yep, I think it's a reasonable feature if you need more speed than decode_slice can give you (e.g. don't want to have the space to write into, or don't want to generate the write bandwidth). If it doesn't autovectorize well (though it has a reasonable shot at doing so), it's pretty amenable to SIMD table lookup ala `pshufb`. 

I think likely the natural home for this is another method on `Engine`.

For `GeneralPurposeEngine`, I'd tackle it like so:

- Partition off the input that's a multiple of 4 tokens in length, leaving the suffix (which might itself be 4 bytes). Then, you can do something like `most_of_the_input.iter().all(|b| table_lookup(b) != INVALID_VALUE)`. This is what will hopefully be autovectorized.
- Since the suffix (1-4 tokens) is where all the complexity is, now you can do the relatively mundane but finicky work of checking the various padding cases per engine config, etc, without worrying about performance.
- Tests. Lots of tests. There should be plenty of decode tests with all sorts of invalid cases to crib from.

PR #287, opened 2025-06-21
Add method to check for decoding errors

tcerqueira:
Feature for #286 

- [x] Add `check_encoded` method to `Engine` trait.
- [ ] Implement feature for all engines.
- [x] Add tests
- [x] Add benchmarks

The tests `decode_invalid_trailing_bytes` are failing because the error for decoding does not match with check implementation. I don't know if it's necessarily incorrect but it's definitely less useful error compared to the decode implementation. Either way, I think the check should error the same way as decode, for everyone's sanity.

The msrv build fails because of range pattern syntax.

### DIFF
--- a/benches/benchmarks.rs
+++ b/benches/benchmarks.rs
@@ -46,6 +46,16 @@ fn do_decode_bench_slice(b: &mut Bencher, &size: &usize) {
     });
 }
 
+fn do_check_encoded(b: &mut Bencher, &size: &usize) {
+    let mut v: Vec<u8> = Vec::with_capacity(size * 3 / 4);
+    fill(&mut v);
+    let encoded = STANDARD.encode(&v);
+
+    b.iter(|| {
+        STANDARD.check_encoded(&encoded).unwrap();
+    });
+}
+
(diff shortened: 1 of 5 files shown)


tcerqueira:
@marshallpierce do you have any suggestion on the best solution to throw the proper error?

tcerqueira:
Benchamarks show it's slower than `decode_slice` 🥀

apasel422:
What's the status of this? I'm interested in the new functionality myself.

tcerqueira:
@apasel422 Stale. Use `decode_slice` and ignore what was written to the buffer.

apasel422:
> @apasel422 Stale. Use `decode_slice` and ignore what was written to the buffer.

Doesn't that require allocating a slice of sufficient size to avoid hitting an error?

marshallpierce:
Sorry I haven't gotten to this; real life has been demanding.

You could use [`DecoderReader`](https://docs.rs/base64/latest/base64/read/struct.DecoderReader.html) to read base64, throwing away the decoded output one byte or chunk at a time. If it succeeds to the end, it's valid, without having to have enough space for the whole thing.

PR #290, opened 2025-11-12
buxfix problems with compiling ron and base64 on nix rustc 1.92.0-nightly (839222065 2025-10-05)

jmikedupont2:
```Compiling base64 v0.22.1 (/data/data/com.termux.nix/files/home/pick-up-nix2/vendor/nix/cargo2nix/vendor/rust-base64)
warning: hiding a lifetime that's elided elsewhere is confusing
  --> src/chunked_encoder.rs:57:26
   |
57 |     pub(crate) fn new(s: &mut String) -> StringSink {
   |                          ^^^^^^^^^^^     ^^^^^^^^^^ the same lifetime is hidden here
   |                          |
   |                          the lifetime is elided here
   |
   = help: the same lifetime is referred to in inconsistent ways, making the signature confusing
   = note: `#[warn(mismatched_lifetime_syntaxes)]` on by default
help: use `'_` for type paths
   |
57 |     pub(crate) fn new(s: &mut String) -> StringSink<'_> {
   |                                                    ++++

warning: `base64` (lib) generated 1 warning (run `cargo fix --lib -p base64` to apply 1 suggestion)````

### CODE REVIEW

File: Cargo.toml [hunk 1]
```diff
@@ -57,3 +57,4 @@ debug = true
 [profile.test]
 # Faster tests save much more than the increase in compilation time
 opt-level = 3
+[workspace]
```
marshallpierce: Why the workspace? Perhaps a stray change from some tooling?

jmikedupont2: yes sorry, ignore that

jmikedupont2:
removed the workspace, also some issues i had were nix building with no warnings for some reason

marshallpierce:
The compiler warning was addressed in #292.

PR #297, opened 2026-05-04
Remove unnecessary clones on tests

FranciscoTGouveia:
As `.clone()` is being called on a fresh `DecodeError::InvalidPadding`, or `DecodeSliceError::OutputSliceTooSmall`, it can be safely removed.

CC @nunoplopes

### DIFF
--- a/src/decode.rs
+++ b/src/decode.rs
@@ -403,7 +403,7 @@ mod coverage_gaming {
 
     #[test]
     fn decode_error() {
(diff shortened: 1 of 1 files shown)


marshallpierce:
The intent of that module is to exercise unimportant code that otherwise would clutter up coverage reports. So, the clone() actually is meaningful, as a way to squelch indications that clone() isn't covered.

PR #299, opened 2026-07-12
feat: add SIMD acceleration

geeknoid:
Add SIMD-accelerated base64 for the STANDARD and URL_SAFE alphabets, controlled
by the `simd-unsafe` feature (on by default). It is the only feature that uses
`unsafe`; disabling it makes the crate `#![forbid(unsafe_code)]`. When enabled,
`unsafe` is confined to the SIMD module and the crate uses `#![deny(unsafe_code)]`.

Three engines are provided, all reusing the scalar `GeneralPurpose` kernels for
the tail and fallback:

- `Simd`: detects the best instruction set at runtime (AVX2 on x86_64, NEON on
  aarch64) and falls back to scalar. Needs `std` for the detection.

- `Avx2` / `Neon`: target one instruction set with no runtime detection, so they
  work in `no_std`. `Avx2` exposes a checked `new` (std) and an
  `unsafe new_unchecked`; NEON is mandatory on aarch64 so `Neon::new` is safe.

- `GeneralPurpose` stays a pure-scalar engine; the encode/decode helpers take an
  optional SIMD-prefix step that is a no-op for it.

Instruction counts below were gathered with gungraun (Callgrind); wall-clock
times with the criterion benchmarks (AVX2, x86-64-v3, this machine):

| Operation    | Input | Instructions (scalar -> SIMD)  | Wall-clock (scalar -> SIMD) |
| ------------ | ----- | ------------------------------ | --------------------------- |
| encode_slice | 3 KiB | 22,245 -> 4,741        (4.7x)  | 1.73 us -> 468 ns   (3.7x)  |
| encode_slice | 3 MiB | 22,282,621 -> 3,802,013 (5.9x) | 1.74 ms -> 456 us   (3.8x)  |
| decode_slice | 3 KiB | 28,620 -> 4,817        (5.9x)  | 1.49 us -> 230 ns   (6.5x)  |
| decode_slice | 3 MiB | 28,705,264 -> 4,063,989 (7.1x) | 1.54 ms -> 216 us   (7.1x)  |

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,9 @@ rstest_reuse = "0.6.0"
 default = ["std"]
 alloc = []
 std = ["alloc"]
+# Enables the SIMD-accelerated engines (`Simd`, `Avx2`, `Neon`). Off by default. It is the only
+# feature that introduces `unsafe`; with it disabled the crate is `#![forbid(unsafe_code)]`.
+simd-unsafe = []
 
 [profile.bench]
 # Useful for better disassembly when using `perf record` and `perf annotate`
--- a/src/engine/general_purpose/decode.rs
+++ b/src/engine/general_purpose/decode.rs
@@ -39,27 +39,97 @@ pub(crate) fn decode_helper(
     decode_table: &[u8; 256],
     decode_allow_trailing_bits: bool,
     padding_mode: DecodePaddingMode,
+    simd_prefix: impl FnOnce(&[u8], usize, &mut [u8]) -> (usize, usize),
 ) -> Result<DecodeMetadata, DecodeSliceError> {
     let input_complete_nonterminal_quads_len =
         complete_quads_len(input, estimate.rem, output.len(), decode_table)?;
 
+    let output_complete_quad_len = input_complete_nonterminal_quads_len / 4 * 3;
+
+    // A SIMD backend, when applicable, decodes a leading prefix; the scalar loops decode the rest.
+    // `simd_prefix` returns `(input_consumed, output_written)` and must consume only whole, valid
+    // quads: `input_consumed % 4 == 0` and `<= input_complete_nonterminal_quads_len` (the terminal
+    // quad is left to `decode_suffix`), `output_written == input_consumed / 4 * 3`, and only those
+    // output bytes are written. It stops on the first invalid/ambiguous quad so the scalar decoder
+    // reports the precise error offset. `(0, 0)` (pure scalar) is always valid.
+    let (input_index, output_index) =
+        simd_prefix(input, input_complete_nonterminal_quads_len, output);
+
+    debug_assert!(input_index % 4 == 0, "prefix must consume whole quads");
+    debug_assert!(
+        input_index <= input_complete_nonterminal_quads_len,
+        "prefix must not consume the terminal quad"
+    );
+    debug_assert!(
+        output_index == input_index / 4 * 3,
+        "prefix output must match consumed input"
+    );
+    debug_assert!(
+        input_index <= input_complete_nonterminal_quads_len,
+        "prefix must not consume the terminal quad"
+    );
+    debug_assert!(
+        output_index == input_index / 4 * 3,
+        "prefix output must match consumed input"
+    );
+
+    decode_complete_quads(
+        input,
+        input_index,
+        input_complete_nonterminal_quads_len,
+        decode_table,
+        output,
+        output_index,
+    )?;
+
+    super::decode_suffix::decode_suffix(
+        input,
+        input_complete_nonterminal_quads_len,
+        output,
+        output_complete_quad_len,
+        decode_table,
+        decode_allow_trailing_bits,
+        padding_mode,
+    )
+}
+
+/// Decode the complete non-terminal quads in `input[input_index_start..input_index_end]` (both
+/// bounds multiples of 4), writing to `output` starting at `output_index_start`, which must equal
+/// `input_index_start / 4 * 3`. Error offsets are reported in absolute input coordinates.
+#[inline]
+fn decode_complete_quads(
+    input: &[u8],
+    input_index_start: usize,
+    input_index_end: usize,
+    decode_table: &[u8; 256],
+    output: &mut [u8],
+    output_index_start: usize,
+) -> Result<(), DecodeSliceError> {
+    debug_assert!(
+        input_index_start % 4 == 0,
+        "quad start must be quad-aligned"
+    );
+    debug_assert!(input_index_end % 4 == 0, "quad end must be quad-aligned");
+    debug_assert!(
+        output_index_start == input_index_start / 4 * 3,
+        "output start must match consumed input"
+    );
+
     const UNROLLED_INPUT_CHUNK_SIZE: usize = 32;
     const UNROLLED_OUTPUT_CHUNK_SIZE: usize = UNROLLED_INPUT_CHUNK_SIZE / 4 * 3;
 
-    let input_complete_quads_after_unrolled_chunks_len =
-        input_complete_nonterminal_quads_len % UNROLLED_INPUT_CHUNK_SIZE;
-
-    let input_unrolled_loop_len =
-        input_complete_nonterminal_quads_len - input_complete_quads_after_unrolled_chunks_len;
+    let quads_len = input_index_end - input_index_start;
+    let unrolled_loop_len = quads_len - quads_len % UNROLLED_INPUT_CHUNK_SIZE;
+    let input_unrolled_loop_end = input_index_start + unrolled_loop_len;
 
     // chunks of 32 bytes
-    for (chunk_index, chunk) in input[..input_unrolled_loop_len]
+    for (chunk_index, chunk) in input[input_index_start..input_unrolled_loop_end]
         .chunks_exact(UNROLLED_INPUT_CHUNK_SIZE)
         .enumerate()
     {
-        let input_index = chunk_index * UNROLLED_INPUT_CHUNK_SIZE;
-        let chunk_output = &mut output[chunk_index * UNROLLED_OUTPUT_CHUNK_SIZE
-            ..(chunk_index + 1) * UNROLLED_OUTPUT_CHUNK_SIZE];
+        let input_index = input_index_start + chunk_index * UNROLLED_INPUT_CHUNK_SIZE;
+        let output_base = output_index_start + chunk_index * UNROLLED_OUTPUT_CHUNK_SIZE;
+        let chunk_output = &mut output[output_base..output_base + UNROLLED_OUTPUT_CHUNK_SIZE];
 
         decode_chunk_8(
             &chunk[0..8],
@@ -88,36 +158,23 @@ pub(crate) fn decode_helper(
     }
 
     // remaining quads, except for the last possibly partial one, as it may have padding
-    let output_unrolled_loop_len = input_unrolled_loop_len / 4 * 3;
-    let output_complete_quad_len = input_complete_nonterminal_quads_len / 4 * 3;
+    let output_after_unroll_start = output_index_start + unrolled_loop_len / 4 * 3;
+    for (chunk_index, chunk) in input[input_unrolled_loop_end..input_index_end]
+        .chunks_exact(4)
+        .enumerate()
     {
-        let output_after_unroll = &mut output[output_unrolled_loop_len..output_complete_quad_len];
+        let output_base = output_after_unroll_start + chunk_index * 3;
+        let chunk_output = &mut output[output_base..output_base + 3];
 
-        for (chunk_index, chunk) in input
-            [input_unrolled_loop_len..input_complete_nonterminal_quads_len]
-            .chunks_exact(4)
-            .enumerate()
-        {
-            let chunk_output = &mut output_after_unroll[chunk_index * 3..chunk_index * 3 + 3];
-
-            decode_chunk_4(
-                chunk,
-                input_unrolled_loop_len + chunk_index * 4,
-                decode_table,
-                chunk_output,
-            )?;
-        }
+        decode_chunk_4(
+            chunk,
+            input_unrolled_loop_end + chunk_index * 4,
+            decode_table,
+            chunk_output,
+        )?;
     }
 
-    super::decode_suffix::decode_suffix(
-        input,
-        input_complete_nonterminal_quads_len,
-        output,
-        output_complete_quad_len,
-        decode_table,
-        decode_allow_trailing_bits,
-        padding_mode,
-    )
+    Ok(())
 }
 
 /// Returns the length of complete quads, except for the last one, even if it is complete.
--- a/src/engine/general_purpose/mod.rs
+++ b/src/engine/general_purpose/mod.rs
@@ -18,7 +18,8 @@ pub(crate) const INVALID_VALUE: u8 = 255;
 
 /// A general-purpose base64 engine.
 ///
-/// - It uses no vector CPU instructions, so it will work on any system.
+/// - It uses no vector CPU instructions, so it will work on any system. For a version that uses
+///   SIMD where available, see the SIMD engines behind the `simd-unsafe` feature.
 /// - It is reasonably fast (~2-3GiB/s).
 /// - It is not constant-time, though, so it is vulnerable to timing side-channel attacks. For loading cryptographic keys, etc, it is suggested to use the forthcoming constant-time implementation.
 
@@ -28,6 +29,12 @@ pub struct GeneralPurpose {
     decode_table: [u8; 256],
     config: GeneralPurposeConfig,
 }
+
+/// A purely scalar base64 engine that never uses hardware-specific vector instructions.
+///
+/// This is an alias for [`GeneralPurpose`], giving an explicit name for callers who want to
+/// guarantee a scalar-only implementation.
+pub type Scalar = GeneralPurpose;
 
 impl GeneralPurpose {
     /// Create a `GeneralPurpose` engine from an [Alphabet].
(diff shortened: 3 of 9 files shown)


geeknoid:
I don't know what the CI failure is about. Since I didn't introduce/change dependencies, the failure seems unrelated to my changes.

Cargo.toml
```diff
@@ -46,9 +46,13 @@ rstest = "0.13.0"
 rstest_reuse = "0.6.0"
 
 [features]
-default = ["std"]
+default = ["std", "simd-unsafe"]
```
marshallpierce: I'd prefer off by default, at least for now. We can always flip it on later.

src/engine/general_purpose/mod.rs
```diff
@@ -11,14 +11,17 @@ use core::convert::TryInto;
 
 pub(crate) mod decode;
 pub(crate) mod decode_suffix;
+#[cfg(feature = "simd-unsafe")]
+pub mod simd;
```
marshallpierce: Any particular reason to have `simd` inside `general_purpose`? It's a separate engine, so I think it warrants its own module under `engine`. I suppose there is an aspect of claiming that `simd` is suitable for "general purpose", but ... eh. Naming is hard.

I considered naming `GeneralPurpose` `Scalar` originally but felt the name was a little confusing to those not indoctrinated in performance.

Maybe with simd finally here it would be good to have a `Scalar` alias at least so that we could offer a choice that promises to never, ever, do unsafe hardware dependent things. But, I digress.

src/engine/general_purpose/mod.rs
```diff
@@ -11,14 +11,17 @@ use core::convert::TryInto;
...
 
 pub use decode::GeneralPurposeEstimate;
 
 pub(crate) const INVALID_VALUE: u8 = 255;
 
 /// A general-purpose base64 engine.
 ///
-/// - It uses no vector CPU instructions, so it will work on any system.
+/// - It uses no vector CPU instructions, so it will work on any system. For a version that uses
+///   SIMD where available, see the SIMD engines behind the `simd-unsafe` feature.
```
marshallpierce, line 22: I wonder, should we leave GP vague so that perhaps in some future time we could feature detect and dispatch to SIMD? In a world where more of SIMD was safe, that might be viable. 

This is a little more sane if we have the aforementioned `Scalar` alias that could be promoted into its own struct some day if needed, for people who want that. There's _an_ argument, though not a _fully_ convincing one, that maybe GeneralPurpose should be "the right choice for most people", in the same way that crypto software uses SIMD under the hood because that's right for most people.

src/engine/general_purpose/decode.rs
```diff
@@ -39,27 +39,66 @@ pub(crate) fn decode_helper(
...
     let input_complete_nonterminal_quads_len =
         complete_quads_len(input, estimate.rem, output.len(), decode_table)?;
 
+    let output_complete_quad_len = input_complete_nonterminal_quads_len / 4 * 3;
+
+    // A SIMD backend, when applicable, decodes a leading prefix; the scalar loops decode the rest.
+    // The last (possibly padded) quad is always left to `decode_suffix`. The prefix consumes whole
+    // quads, so `input_index` and `output_index` stay aligned for the scalar loops.
+    let (input_index, output_index) =
+        simd_prefix(input, input_complete_nonterminal_quads_len, output);
```
marshallpierce, line 56: this would be a good place for a debug_assert! to ensure that the contract of simd_prefix is obeyed (in code hit by tests, anyway)

src/engine/general_purpose/decode.rs
```diff
@@ -39,27 +39,66 @@ pub(crate) fn decode_helper(
     decode_table: &[u8; 256],
     decode_allow_trailing_bits: bool,
     padding_mode: DecodePaddingMode,
+    simd_prefix: impl FnOnce(&[u8], usize, &mut [u8]) -> (usize, usize),
```
marshallpierce, line 42: document the contract for simd_prefix. Or, perhaps, it should be a trait so that the contract has one obvious place to be documented?

src/engine/general_purpose/decode.rs
```diff
@@ -39,27 +39,66 @@ pub(crate) fn decode_helper(
...
+        output_complete_quad_len,
+        decode_table,
+        decode_allow_trailing_bits,
+        padding_mode,
+    )
+}
+
+/// Decode the complete non-terminal quads in `input[input_index_start..input_index_end]` (both
+/// bounds multiples of 4), writing to `output` starting at `output_index_start`, which must equal
+/// `input_index_start / 4 * 3`. Error offsets are reported in absolute input coordinates.
```
marshallpierce, line 98: these conditions also feel like good uses of `debug_assert!`

src/engine/general_purpose/decode.rs
marshallpierce, line 42: The addition of a (sugared away) generic might monomorphize into a surprising level of code growth, so it would be worth checking how much it adds to a binary, especially if multiple engines are in use. Might need to do the mitigation that Engine#encode does.

src/engine/general_purpose/mod.rs
```diff
@@ -192,6 +81,133 @@ impl super::Engine for GeneralPurpose {
...
+pub(crate) fn encode_helper(
+    encode_table: &[u8; 64],
+    input: &[u8],
+    output: &mut [u8],
+    simd_prefix: impl FnOnce(&[u8], &mut [u8]) -> (usize, usize),
+) -> usize {
+    const BLOCKS_PER_FAST_LOOP: usize = 4;
+    const LOW_SIX_BITS: u64 = 0x3F;
+
+    let (mut input_index, mut output_index) = simd_prefix(input, output);
```
marshallpierce: Since I'm apparently in a `debug_assert!` mood, how about here too

src/engine/simd.rs
```diff
@@ -0,0 +1,776 @@
...
+//!
+//! Only the STANDARD and URL_SAFE alphabets are accelerated (they share indices `0..=61` and differ
+//! only at `62`/`63`); any other alphabet falls back to the scalar path.
+//!
+//! The kernels follow Wojciech Muła's vectorized base64 algorithms
+//! (<http://0x80.pl/notesen/2016-01-17-sse-base64-decoding.html> and the companion encoding note).
+//! AVX2 uses the multiply-based bit (de)interleave; NEON, which lacks the relevant multiplies, uses
+//! the shift/mask variant. Both share the same per-alphabet lookup tables.
+#![allow(unsafe_code)]
+// NEON intrinsics require Rust 1.59, above the crate MSRV; these paths need `std`/the feature and
```
marshallpierce, line 22: This seems like a good reason to me to bump MSRV. I'm of a mind to jump to 1.85 for edition 2024 when this crate finally goes 1.0 anyway so we aren't stuck in the dark ages forever.

src/engine/general_purpose/simd.rs
```diff
@@ -0,0 +1,776 @@
...
+    inner: GeneralPurpose,
+    kind: SimdKind,
+    backend: Backend,
+}
+
+#[cfg(all(feature = "std", any(target_arch = "x86_64", target_arch = "aarch64")))]
+impl Simd {
+    /// Create a `Simd` engine, detecting the available instruction set once.
+    #[must_use]
+    pub fn new(alphabet: &Alphabet, config: GeneralPurposeConfig) -> Self {
```
marshallpierce: I think it would be fine to only offer constructors for the standard and url safe alphabets, and not bother detecting / fallback for others. 

On the other hand, maybe if we later do something like generating tables at runtime for a pure pshufb algorithm, then we'd want to support all alphabets...

src/engine/general_purpose/simd.rs
```diff
@@ -0,0 +1,776 @@
...
+        },
+        DecodeMetadata, Engine,
+    },
+    DecodeSliceError,
+};
+
+/// Which accelerated alphabet family an alphabet belongs to, if any.
+#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum SimdKind {
```
marshallpierce: If we got rid of `Other`, would it help performance at all to be able to express the tables as a trait w/ impls for Standard and UrlSafe so that the table lookups could be more aggressively inlined?

src/engine/simd.rs
```diff
@@ -0,0 +1,776 @@
...
+        input: &[u8],
+        output: &mut [u8],
+    ) -> (usize, usize) {
+        let lut_data = match encode_lut(kind) {
+            Some(lut) if input.len() >= SIMD_MIN_INPUT_ENCODE => lut,
+            _ => return (0, 0),
+        };
+        let lut = _mm256_broadcastsi128_si256(_mm_loadu_si128(lut_data.as_ptr().cast()));
+        #[rustfmt::skip]
+        let shuf = _mm256_setr_epi8(
```
marshallpierce, line 133: All unsafe calls will need a `// SAFETY` comment justifying why they're safe, or this won't pass the unsafe review at {employer}. :/

marshallpierce (changes requested):
More to come as I pore over simd.rs, but here's what I've got so far. 

BTW, I've fixed the CI issue on `master` (by updating the toolchain expected to be able to build all targets... sigh). Also I've started poking at the relevant people to see if we can get this onboarded into oss-fuzz after all, as it'll be newly relevant with unsafe code.

I'm also curious if you have any feedback on the design of the `Engine` trait, as you've probably interacted with it more than any other contributor in a while. It was intended to enable exactly this sort of addition.

Oh, also -- it might be enjoyable to add {encode|decode}_slice benchmarks using `Simd` so it can be easily compared.

geeknoid:
Thanks for the thorough pass - this was really useful. I've addressed everything below; a
summary of the bigger structural changes:

- Moved the SIMD code out of `general_purpose` into its own `engine::simd` module.
- Flipped `simd-unsafe` off by default.
- Dropped the `Other` alphabet fallback and reworked the per-alphabet tables into a small
  sealed `SimdAlphabet` trait (assoc-const LUTs); the engines now have dedicated
  `::standard()` / `::url_safe()` constructors instead of taking an `&Alphabet`.
- Added `// SAFETY:` justifications on every unsafe pointer op, plus `debug_assert!`s and a
  single documented contract for the scalar/SIMD hand-off.
- Checked the monomorphization cost and factored the scalar encode tail out so it's emitted
  once rather than per engine.
- Added `encode_slice`/`decode_slice` benches for the `Simd` engine.

Cargo.toml
geeknoid: Done, although I worry this is a mistake.

We have a lot of teams using base64, and I'm hoping that they just get a magical perf boost by my changes. But by making the default be off, noone will get this magical perf boost. They would have to discover than one of their dependencies now needs a new feature enabled in order to fully use the hardware capabilities. That's probably never going to happen.

geeknoid:
Overall the engine  trait made this addition genuinely pleasant - I could add three engines that
reuse the scalar kernels without touching the trait itself, which is exactly the extension
point you intended, so I'd call the core design validated.

Two small bits of friction I hit, none blocking:

1. **No seam to reuse another engine's scalar core.** The SIMD engines want to delegate the
   tail (and, for decode, the terminal quad + exact error-offset reporting) to the
   `GeneralPurpose` kernels, but its `encode_table`/`decode_table`/config are private. I
   added `pub(crate)` accessors and threaded an optional "SIMD prefix" closure through
   `encode_helper`/`decode_helper`. That works, but a blessed "here's the scalar core, run a
   vector prefix then hand back to it" seam would make future accelerated engines cheaper and
   keep the error-reporting semantics in one place.

2. **`internal_decode`'s error-offset contract is the hard part to re-implement.** Producing
   byte-exact `InvalidByte`/`InvalidLastSymbol` offsets from a vectorized loop isn't
   practical, so the SIMD path only consumes whole *valid* quads and bails to scalar on the
   first suspect block. The trait accommodated that fine; just noting that the interesting
   surface area for an engine author is the decode error semantics, not the happy path.

Nothing here argues for a trait change right now - mostly "an internal reuse seam would be
nice."

src/engine/general_purpose/mod.rs
geeknoid: Agreed it's a separate engine - moved it to `src/engine/simd.rs` (`engine::simd`). To reach
the scalar tables/config from the sibling module I added a couple of `pub(crate)` accessors on
`GeneralPurpose`/`GeneralPurposeConfig`.

I also added the `Scalar` alias you mentioned:

```rust
pub type Scalar = GeneralPurpose;
```

re-exported from `engine`, so there's an explicit name for "never touches hardware-specific
vector instructions." Left it as an alias rather than a distinct type so it can be promoted
later if needed.

src/engine/general_purpose/mod.rs
geeknoid: I kept `GeneralPurpose` strictly pure-scalar for now: with `simd-unsafe` off the crate stays
`#![forbid(unsafe_code)]`, and having `GeneralPurpose` feature-detect into `unsafe` SIMD would
break that promise for the exact people who pick it for safety. The new `Scalar` alias pins
down "always scalar," and `Simd` is the opt-in "right for most people, uses SIMD under the
hood" choice. If more of this becomes safe later, `GeneralPurpose` dispatching is still open.

src/engine/general_purpose/decode.rs
geeknoid: Added. After the prefix runs, `decode_helper` now asserts it consumed whole quads, didn't
touch the terminal quad, and that output stayed in lockstep:

```rust
debug_assert!(input_index % 4 == 0, "prefix must consume whole quads");
debug_assert!(input_index <= input_complete_nonterminal_quads_len, "prefix must not consume the terminal quad");
debug_assert!(output_index == input_index / 4 * 3, "prefix output must match consumed input");
```

src/engine/general_purpose/decode.rs
geeknoid: Documented the contract in one place - a block comment on the `simd_prefix` parameter in
`decode_helper` (and the equivalent in `encode_helper`) that spells out the alignment/bounds
guarantees and that returning `(0, 0)` is always valid.

I did try the trait version, but a blanket `impl<F: FnOnce(..)> SimdPrefix for F` trips
higher-ranked-lifetime inference on the unannotated no-op/closure call sites ("implementation
of `FnOnce` is not general enough"). Rather than force closure signature annotations
everywhere, I kept the `impl FnOnce` parameter with the single-source-of-truth doc comment.
Happy to revisit if you'd prefer a named wrapper type.

src/engine/general_purpose/decode.rs
geeknoid: Added at the top of `decode_complete_quads`:

```rust
debug_assert!(input_index_start % 4 == 0, "quad start must be quad-aligned");
debug_assert!(input_index_end % 4 == 0, "quad end must be quad-aligned");
debug_assert!(output_index_start == input_index_start / 4 * 3, "output start must match consumed input");
```

src/engine/general_purpose/decode.rs
geeknoid: Checked. Decode was already fine - `decode_helper`'s body is just calls into the non-generic
`decode_complete_quads` / `decode_suffix`, so only a tiny glue wrapper is monomorphized per
engine (measured `internal_decode` at ~0x140-0x190 each, sharing one `decode_complete_quads`).

Encode *did* duplicate: the whole unrolled scalar loop lived inline in the generic
`encode_helper`, so it was emitted once per engine (~1.4 KB x {GeneralPurpose, Simd, Avx2},
plus Neon on aarch64). I applied the same treatment - pulled the loop into a non-generic
`encode_scalar_tail`; the generic wrapper is now just "run prefix, assert, call tail." After:

| symbol | before | after |
| ------ | ------ | ----- |
| `GeneralPurpose::internal_encode` | 0x560 | 0x017 |
| `Avx2::internal_encode`           | 0x594 | 0x076 |
| `Simd::internal_encode`           | 0x5fc | 0x099 |
| `encode_scalar_tail` (shared)     | -     | 0x54e (once) |

So ~2.7 KB of duplicated scalar tail removed, and it now mirrors how decode already shares its
core. (Measured with `nm --print-size` on a release binary exercising all engines.)

src/engine/general_purpose/mod.rs
geeknoid: Added - the encode side asserts the multiple-of-3 / `out == in/3*4` / in-bounds invariants
before the scalar tail runs.

src/engine/simd.rs
geeknoid: Agree in principle, and I'd rather not hold this PR on it. Proposal: land SIMD with the
current MSRV (the NEON paths keep the localized `#![allow(clippy::incompatible_msrv)]` and
aren't built at MSRV), and do the jump to 1.85 / edition 2024 as its own PR alongside the 1.0
work you mentioned. If you'd rather bundle it here, I'm happy to, just say the word.

src/engine/general_purpose/simd.rs
geeknoid: Done. Removed the `Other` classification and the silent scalar fallback entirely. The engines
no longer take an `&Alphabet`; they have dedicated constructors:

```rust
Simd::standard(config) / Simd::url_safe(config)
Avx2::standard(config) / Avx2::url_safe(config)          // Option, CPU-detected
Avx2::standard_unchecked(config) / ..._unchecked(config) // unsafe, no_std
Neon::standard(config) / Neon::url_safe(config)
```

If we later grow a runtime-table pshufb path for arbitrary alphabets, that can come back as an
additional constructor without disturbing these.

src/engine/general_purpose/simd.rs
geeknoid: Did exactly this. There's now a sealed `SimdAlphabet` trait with the per-alphabet LUTs as
associated consts, implemented by `Standard`/`UrlSafe` marker types; the kernels are generic
over the family (`encode_bulk::<A>` / `decode_bulk::<A>`) so the tables are compile-time
constants instead of runtime-selected pointers. A 2-variant `SimdKind` picks the
monomorphization at the dispatch boundary. Correctness verified against the scalar oracle on
both AVX2 and NEON (the latter under QEMU via `cross`).

src/engine/simd.rs
geeknoid: Whoops, I'm used to having this caught by lints.

Fixed.

Cargo.toml
marshallpierce: OK, you've convinced me. I'll make it default to on.

marshallpierce:
Thanks very much for this. I'll merge as is and do a little bit of tidying over the next couple of days, along with incorporating some other pending changes, and then do a release.

marshallpierce:
Released in 0.23.0.

PR #310, opened 2026-08-04
Fix testing on architectures without unsafe SIMD support

### LINKED ISSUE #309: Unable to build and run tests on architectures without unsafe SIMD support
# Before you file an issue

- Did you read the docs? *yes*
- Did you read the README? *yes*

# The problem

Working on any architecture other than `x86_64` or `aarch64`, such as `i686`, `ppc64le`, or `s390x`:

```
$ git clone https://github.com/marshallpierce/rust-base64.git
$ cd rust-base64
$ cargo build
[…]
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 10.00s
$ cargo test
[…]
error[E0432]: unresolved import `crate::engine::simd`
  --> src/engine/tests.rs:6:20
   |
 6 | use crate::engine::simd::Simd;
   |                    ^^^^ could not find `simd` in `engine`
   |
note: found an item that was configured out
  --> src/engine/mod.rs:24:9
   |
19 |       any(
   |  ________-
20 | |         target_arch = "x86_64",
21 | |         all(target_arch = "aarch64", target_feature = "neon")
22 | |     )
   | |_____- the item is gated here
23 |   ))]
24 |   pub mod simd;
   |           ^^^^

```

(For the above, I used an `qemu-user-static`-emulated Fedora Rawhide `ppc64le` [`mock` chroot](https://github.com/rpm-software-management/mock). I first observed this issue while working on updating Fedora’s [`rust-base64` package](https://src.fedoraproject.org/rpms/rust-base64).)

# How I, the issue filer, am going to help solve it

- I will open a PR that improves conditionalization in `src/engine/tests.rs`.

musicinmybrain:
Fixes #309 by improving conditionals within `src/engine/tests.rs`.

This extends the basic approach from

https://github.com/marshallpierce/rust-base64/blob/e34f9a08c5c89a4641350ac22033f3fa4f5d4d97/src/engine/mod.rs#L17-L24

throughout the test code as needed to support testing on all architectures.

Particuarly, this ensures that the case where the `simd-unsafe` feature is enabled (as when default features are used), but  no unsafe SIMD implementation exists for the target, is always handled.

### DIFF
--- a/src/engine/tests.rs
+++ b/src/engine/tests.rs
@@ -2,7 +2,13 @@
 #![allow(unused_variables)]
 
 use crate::alphabet::Symbol;
-#[cfg(feature = "simd-unsafe")]
+#[cfg(all(
+    feature = "simd-unsafe",
+    any(
+        target_arch = "x86_64",
+        all(target_arch = "aarch64", target_feature = "neon")
+    )
+))]
 use crate::engine::simd::Simd;
 #[cfg(all(
     feature = "simd-unsafe",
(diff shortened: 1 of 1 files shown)


marshallpierce:
Sorry for the breakage, and thanks for the fix. I'll release this promptly, and after that we can figure out how to incorporate this into CI. Do you have suggestions on a simple way to check that `cargo test` works on less common architectures that won't burn a ton of CI minutes?

decathorpe:
Using `miri` to run tests for those architectures might work? At the very least it should be less setup overhead for CI than full emulation.

PR #312, opened 2026-09-14
Expand Base64 performance benchmark coverage

optiklab:
## Problem

The existing Criterion suite samples only a few input sizes and groups unrelated APIs together, which makes small-input and SIMD-crossover analysis difficult. Decode throughput labels also do not distinguish raw data size from encoded input size.

## Change

Expand the benchmark matrix to cover these raw input sizes:

```text
0, 1, 2, 3, 4, 8, 16, 24, 32, 48, 64, 96,
128, 192, 256, 384, 512, 768, 1024,
4096, 16384, 65536, 1048576
```

The 3 MiB, 10 MiB, and 30 MiB cases remain in a separate very-large group.

The new groups separate:

- latency and throughput sizes;
- scalar `GeneralPurpose`, runtime `Simd`, and explicit AVX2 or NEON engines;
- allocating, slice-output, and reuse-buffer APIs;
- standard and URL-safe alphabets;
- padded and unpadded configurations;
- valid decode inputs and invalid beginning, middle, end, trailing-bit, padding, and SIMD-tail cases.

Inputs and reusable output buffers are created outside timed loops. Encode throughput uses raw input bytes. Decode benchmark names include raw and encoded sizes, and decode throughput uses encoded input bytes. Input generation is deterministic for comparable runs.

`benches/README.md` documents the group layout, focused commands, and Criterion 0.7 baseline workflow. For example:

```bash
cargo bench --bench benchmarks -- \
  "encode/latency/standard_padded/general_purpose/slice_raw_bytes/64" --exact
cargo bench --bench benchmarks -- \
  "decode/errors/standard_padded/avx2/invalid_end"
```

This PR changes measurement coverage, not codec performance.

## Correctness and safety

- The codec implementation is unchanged.
- `cargo test --benches --all-features` executed every registered benchmark setup and invalid-input assertion successfully on x86-64, including AVX2.
- Scalar-only and ARM64 NEON registration paths were type-checked.
- No unsafe code is added.
- `no_std`, alloc-only, and cross-target builds pass.
- MSRV and public APIs are unchanged.

## Benchmark method

- baseline commit: `5b98ee180cc7c2a64af6cb12c724b1721550e0d0`
- candidate commit: `a073c48c80c41a8705b757f61bdbeb5eb3a02053`
- CPU and architecture: Intel Xeon Platinum 8370C, x86-64
- OS: Windows 11 Enterprise 10.0.26200
- rustc: 1.98.0 (`88d9e12ae`, LLVM 22.1.8)
- target features: native runtime AVX2 detection; no additional `RUSTFLAGS`
- build flags: Criterion bench profile, optimized with debug information
- benchmark command: focused commands shown above with `--quick --noplot` for execution smoke checks
- repetitions and ordering: not applicable; no parent/candidate codec comparison was made
- statistic reported: none; smoke results validate execution only

## Regressions and tradeoffs

- The full matrix contains substantially more benchmarks, so running an unfiltered suite takes longer. Predictable substring filters and a separate low-sample very-large group keep focused work practical.
- Explicit AVX2 benchmarks are omitted on CPUs without AVX2 runtime support.
- NEON execution was not measured on my x86-64 host; its registration path was cross-compiled for `aarch64-pc-windows-msvc`.
- Streaming decoder, string-writer, and display benchmarks are no longer part of this focused codec-engine matrix. `EncoderWriter<Vec<u8>>` remains covered as an append/reuse API.

## Validation

```text
cargo fmt -- --check
cargo clippy --all-targets --all-features -- -D warnings
cargo test --no-default-features
cargo test --no-default-features --features alloc
cargo test --all-features
cargo test --benches --all-features
cargo build --all-targets --all-features
cargo build --target thumbv6m-none-eabi --no-default-features
cargo build --target thumbv6m-none-eabi --no-default-features --features alloc
cargo build --target wasm32-unknown-unknown --no-default-features
cargo check --bench benchmarks --no-default-features --features std
cargo check --target aarch64-pc-windows-msvc --bench benchmarks --all-features
cargo bench --bench benchmarks --no-run
```


marshallpierce:
Thanks for the contributions. I'll get to this and the following PRs in the next week or two, hopefully.

